PR #4, opened 2017-10-10
Config File Support

nocalla:
From https://github.com/nocalla/transaction_csv_cleanup. Properly pasted this time!


nocalla:
Implemented configuration file support. Try this out and see what you think!

torbengb:
I just went ahead and merged this. If you can do it yourself, feel free to do so! If you can't, I need to figure out how to give you that ability!

nocalla:
I don't think I can!

torbengb:
You can't? Then I'm really sorry that I didn't do this for you several DAYS ago already. Apologies! I'll look into shared permissions.

I ran the new .py file and got the error:
    ImportError: No module named configparser

A minute of googling told me how to install the missing module with the linux command:
    sudo python -m pip install configparser
We'll need to add a section to the readme (or somewhere else) about required modules, and how to install them.

That seemed to do the trick, except obviously I haven't modified the config file to my needs so it failed on the find_directory(). We need a section in the readme about that, too. Not saying YOU need to write it though.

torbengb:
I've added you as a "collaborator". Not sure what that means, haven't got time to RTFM right now but perhaps that gives you access?

nocalla:
Oh! I didn't have to install it, that's a bit strange. We can do a local
Python install that will have the required packages. Collaborator seems to
have given me the correct access. Thanks!

PR #7, opened 2017-10-17
Torbengb patch 1

torbengb:
Line 58: I'd prefer to print an error message rather than silently skipping it, but I couldn't make it work yet so for now I'll just share the part that works :-)

Line 70: Fixed TypeError: "delimiter" must be string, not unicode

Line 111 **unsolved**: 'newline' is not valid in Linux, this line only works without newline = "". Need to solve that.

### DIFF
--- a/transaction_csv_cleanup.py
+++ b/transaction_csv_cleanup.py
@@ -55,7 +55,7 @@ def get_files():
     b = g_config["input_filename"]
     c = g_config["fixed_prefix"]
     
-    if b is not "":
+    if b is not "" and g_config["path"] is not "":
         os.chdir(find_directory())
         return [f for f in os.listdir(".") if f.endswith(a) if b in f if c not in f]
     return []
--- a/transaction_csv_cleanup.py
+++ b/transaction_csv_cleanup.py
@@ -23,7 +23,7 @@ def get_configs():
     # get all our config files
     conf_files = [f for f in os.listdir(".") if f.endswith(".conf")]
     config = configparser.ConfigParser()
-    config.read(conf_files)
+    config.read(conf_files, encoding = "utf-8")
     return config
     
 def fix_conf_params(section): # to do
@@ -145,7 +145,6 @@ def main():
         g_config = fix_conf_params(all_configs[section])
         # find all applicable files
         files = get_files()
-        print("{}, {}".format(section, files))
         for file in files:
             # create cleaned csv for each file
             output = clean_data(file)
--- a/transaction_csv_cleanup.py
+++ b/transaction_csv_cleanup.py
@@ -67,7 +67,7 @@ def clean_data(file):
     has_headers = g_config["has_headers"]
     output_data = []
     with open(file) as transaction_file:
-        transaction_reader = csv.reader(transaction_file, delimiter = delim)
+        transaction_reader = csv.reader(transaction_file, delimiter=str(delim).encode('utf-8'), quotechar=str(u'"').encode('utf-8'))
         transaction_data = list(transaction_reader)
 
         # make each row of our new transaction file
--- a/transaction_csv_cleanup.py
+++ b/transaction_csv_cleanup.py
@@ -1,0 +1,158 @@
+#! /usr/bin/python
+# transaction_csv_cleanup.py
+# for Python 3
+
+# Searches specified folder or default download folder for exported
+# bank transaction file (.csv format) & adjusts format for YNAB import
+
+# OPERATIONS
+#   ~ Find all .conf files & get all configuration within
+#   ~ Do following for each configuration found
+#   ~ Find & open TransactionExport.csv for processing
+#   ~ Change columns from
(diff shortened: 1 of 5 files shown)


nocalla:
Line 58: I don't understand why you're checking if the path is "". If it's not there, the script automatically defaults to finding the downloads directory. What do you want it to do there?
Line 70: This is a bit messy - it offends my sense of style! But yes, I'll see if I can switch encoding at an input level rather than having to do it in the function.
Line 111: I can't find any documentation stating this! Is it definitely a Linux issue?

transaction_csv_cleanup.py
```diff
@@ -55,7 +55,7 @@ def get_files():
     b = g_config["input_filename"]
     c = g_config["fixed_prefix"]
     
-    if b is not "":
+    if b is not "" and g_config["path"] is not "":
```
nocalla: I don't understand what you want to happen here.

transaction_csv_cleanup.py
```diff
@@ -67,7 +67,7 @@ def clean_data(file):
     has_headers = g_config["has_headers"]
     output_data = []
     with open(file) as transaction_file:
-        transaction_reader = csv.reader(transaction_file, delimiter = delim)
+        transaction_reader = csv.reader(transaction_file, delimiter=str(delim).encode('utf-8'), quotechar=str(u'"').encode('utf-8'))
```
nocalla: Messy. I think I'd rather encode things when getting them from the config file. What's the quotechar section?

nocalla: Also...I'm not getting this error? From a quick Google it seems to be a Python 2.x issue.

transaction_csv_cleanup.py
torbengb: This is meant to avoid an ugly error when no path is defined.

nocalla:
But if no path is defined, it defaults to the user's download directory
using Lines 120-131. What error do you get?

transaction_csv_cleanup.py
torbengb: Py2.x ... this is so odd! I must try to uninstall/purge python from my system and make an explicit py3 installation. Let's see how much of this goes away then. I'll respond with results! Let's keep this unmerged until then. 

I agree that it's probably messy. I was trying to make it work using minimal changes; a more elegant solution would probably belong somewhere else in the code. My system was complaining that the delimiter character wasn't in Unicode, so this is what I found as a solution to convert it to U. The quotechar was part of what I found; it's also a parameter like the delimiter and it reminded me that things like Excel also allow me to specify something other than the typical double quote as the quotation mark, for instance just a single quote instead. If we keep it, it should be treated elegantly like the unicode-ification of the delimiter character, and also included in the conf file.

torbengb:
When the 'Source Path' is blank in the config file, I get this output:

      File "./transaction_csv_cleanup.py", line 60, in get_files
        os.chdir(find_directory())
    OSError: [Errno 2] No such file or directory: ''

torbengb:
Oh f'ck, I just removed Python from my system and it took ~30 applications with it because they depend on Python, total about 1GB of files! :-) Looks like I will have a little bit of cleanup to do --- tomorrow!!

transaction_csv_cleanup.py
nocalla: I think it's straightforward to import as unicode. Think configparser has an encode function that can do this.

nocalla:
I'm off work tomorrow so I'll take a proper look at these issues.

transaction_csv_cleanup.py
nocalla: Fixed this one by modifying line 26 to 
`    config.read(conf_files, encoding = "utf-8")
`
Will commit to mine later. Don't think I can modify a pull request. Or maybe I can...

nocalla: I can! I cut out the quotechar bit for now, but will look at it again when the base stuff is actually working cross platform!

transaction_csv_cleanup.py
```diff
@@ -145,7 +145,6 @@ def main():
         g_config = fix_conf_params(all_configs[section])
         # find all applicable files
         files = get_files()
-        print("{}, {}".format(section, files))
```
nocalla: A stray debugging line!

torbengb:
Thanks for the discussion, I enjoy working with you. I'll retract this request and make a new one later (when my machine is really, truly, running Py3 :-)

nocalla:
Was it python 2 so?

On 18 Oct 2017 11:09, "Torben Gundtofte-Bruun" <notifications@github.com>
wrote:

Thanks for the discussion, I enjoy working with you. I'll retract this
request and make a new one later (when my machine is really, truly, running
Py3 :-)

—
You are receiving this because your review was requested.

Reply to this email directly, view it on GitHub
<https://github.com/torbengb/bank2ynab/pull/7#issuecomment-337537128>, or mute
the thread

PR #11, opened 2017-10-18
changed first line to python3; added screen output.

torbengb:
I have changed the first line from '#! /usr/bin/python' to '#! /usr/bin/python3' instead, to explicitly have my system use the Python version 3. As you expected, that cleared up all of the issues! 

Two questions:

Can we keep 'python3' or does that break anything on your side?

My testing showed that the script stops at line 71 'transaction_data = list(transaction_reader)' if either the file ends with a newline, or the file contains special characters like üöäÜÖÄßæøåÆØÅ. Shouldn't the whole Unicode business already take care of umlauts?

### DIFF
--- a/transaction_csv_cleanup.py
+++ b/transaction_csv_cleanup.py
@@ -1,4 +1,4 @@
-#! /usr/bin/python
+#! /usr/bin/python3
 # transaction_csv_cleanup.py
 # for Python 3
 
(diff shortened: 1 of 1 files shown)


nocalla:
That's interesting about the special characters. Unicode stuff always confuses me. Printing filenames can be problematic if the console can't handle Unicode too.

nocalla:
Make the special characters an issue I'd say. We'll have to investigate.

Issue #25, opened 2017-10-19
Python 2.x Support


nocalla:
As requested by [u/fio_smiles](https://www.reddit.com/r/ynab/comments/77d7rt/tools_bank2ynab_heres_a_script_that_helps_with/dol9p0u/), this may be possible. Need handling for the following main issues:

- [x] ConfigParser to configparser
- [x] csv delimiter as string, not unicode
- [x] csv newline invalid

Assigning myself to this issue. It's not a high importance item, but might help prevent problems if @torbengb starts accidentally using Python 2 again...

torbengb:
As a relevant side-note and reference, [here's why Python 2 is still default on Linux](https://askubuntu.com/a/590030/5786) even though 3.x is factory-installed as well. 
TLDR: A ton of system programs use Python and haven't been ported to Python 3 yet. Any attempt to force the system default to Python 3 would cause big trouble outside of `bank2ynab`.

toyg:
I've pushed a separate py2 branch at 98c0a17cc583d52f9a78af9c8f0befa6bd76ff3e as a playground for this work. I took care of differences in stdlib name and interfaces, and now it seems to run fine with both 2 and 3. Tested with Python 2.7.10 and 3.6.1 on OSX, would be good if somebody could test on Windows before it's merged, because one of the renamed modules is `winreg`; and ideally with python < 3.6 (although I don't expect any breakage on 3.4+).

Not sure what you mean by "csv delimiter as string, not unicode", because `csv` seems to take care of business without any particular handling...?

nocalla:
It's referenced in more detail over on the py2 branch at #26. Will test out on Windows and merge into the py2 testing branch.

torbengb:
I can also test on Linux tonight = v2.7 and 3.5.1

nocalla:
Issues listed probably fixed in latest master push. Closing to clean up issues. We can reopen if need be.

PR #26, opened 2017-10-20
code changes via PM from u/fio_smiles

torbengb:
code changes via PM from u/fio_smiles to address issue #25: "I don't know if that will work with Python 3+". Let's review and test it, see if we can build on it.
I will test it on my Linux tonight.


nocalla:
PM him back and get him to join Github so we can credit him properly!

nocalla:
Or her! Accidental sexism, sorry!

torbengb:
I'm working on it via PM :-) S/he is already intrigued by GitHub and might sign up. Would be cool!

torbengb:
Okay I'm having trouble running this modified code. As-is I got this error: 

      File "newtest.py", line 27, in <module>
        from future import unicode_literals # issue #25
    ImportError: No module named future

So I tried installing `future` with the usual command `sudo python -m pip install future` but then it said:

    Could not import setuptools which is required to install from a source distribution.
    Please install setuptools.

So I did: `sudo python -m pip install setuptools` followed by `sudo python -m pip install future`. Then I got: 

      File "newtest.py", line 27, in <module>
        from future import unicode_literals # issue #25
    ImportError: cannot import name unicode_literals

I don't know what `unicode_literals` is but I tried installing that like above, but failed ("No matching distribution found for unicode_literals").

Without a better understanding of Python, I'm going to have to pass on this code change. Sorry.

nocalla:
I'll see if I can work it out over the weekend.

fiosmiles:
She :) 
Not a developer and just hacked about with this to make it work, but I'll take a look next week and see if I can help.

It's __future__ which may have been lost in formatting, see this link:
http://python-future.org/unicode_literals.html

fiosmiles:
Ahhh, sorry the automatically formating is ruining it, there's two underscores before and after the word future, if you follow the link it should make more sense.

torbengb:
Welcome @fiosmiles!! Great to see you on board! 

I've added the underscores (see new commit linked above) and now there's just a single issue left, but I can't figure out how to solve it:

      File "bank2ynab.py", line 81, in clean_data
        transaction_reader = csv.reader(transaction_file, delimiter = delim)
    TypeError: "delimiter" must be string, not unicode
 
I've tried appending `.encode('utf-8')` or `.decode('utf-8')` but it doesn't seem to make a difference. 

That's when I force the script to run as Python 2 (using the linux command `python2 bank2ynab.py`). When I run it as Python 3 (`python3 bank2ynab.py`), it works fine.

nocalla:
I haven't played with this yet. That __future__ thing is interesting! 

Does the CSV without the newline output correctly on python 3? The reason I had it was because I was getting extra blank rows between each row without.

torbengb:
> Does the CSV without the newline output correctly on python 3? 

Yes it does. I think the `"wb"` setting in line 125 did the trick.

nocalla:
I think I ran into trouble with that on Windows. Will report back later.

nocalla:
I'm getting errors. 
For the write_data change to "wb":
`Traceback (most recent call last):
  File "bank2ynab.py", line 174, in <module>
    main()
  File "bank2ynab.py", line 167, in main
    write_data(file, output)
  File "bank2ynab.py", line 129, in write_data
    writer.writerow(row)
TypeError: a bytes-like object is required, not 'str'`

And for encoding the delimiter:
`Traceback (most recent call last):
  File "bank2ynab.py", line 174, in <module>
    main()
  File "bank2ynab.py", line 166, in main
    output = clean_data(file)
  File "bank2ynab.py", line 84, in clean_data
    transaction_reader = csv.reader(transaction_file, delimiter = delim)
TypeError: "delimiter" must be string, not bytes`

fiosmiles:
@nocalla is there a sample file that everyone is using?  If so, can you point me at it and I'll try that file, or if not can you share your file or some version of it and I'll try it on my end.

fiosmiles:
Also, this is something I was looking at when I hacked this to work for me on 2.7, maybe the switch is worth implementing if it's just not compatible with 3+

https://www.reddit.com/r/ynab/comments/77d7rt/tools_bank2ynab_heres_a_script_that_helps_with/

fiosmiles:
Wrong link, correction: https://stackoverflow.com/questions/29840849/writing-a-csv-file-in-python-that-works-for-both-python-2-7-and-python-3-3-in

nocalla:
There isn't a test file. We should definitely make one. That has umlauts
and ends with a newline! I'll have a read of the link and see what I can
make of it if I can work out what's going wrong on my end.

torbengb:
Here's a test file, feel free to add more in that folder: https://github.com/torbengb/bank2ynab/tree/master/test-data 
I hope I anonymized it sufficiently :-)

nocalla:
I like [this](https://stackoverflow.com/a/29841468) solution for the CSV issue.

nocalla:
Also, we have to account for the renamed configparser! I think both of you have installed the Python 3 version as a workaround, but we should make it dependent on the pre-installed modules only. I think the best way to do that is import ConfigParser as configparser near the top of the code!

nocalla:
This all works fine at my end, but I need @fiosmiles and @torbengb to test it out, because I **know** it won't work properly at your end! If we can pinpoint the errors, then we can track down what to focus on, so please paste your lovely exceptions.

torbengb:
I'll test your code on my Linux (tonight?) and reply then.

torbengb:
Sorry to bring bad news, but it doesn't run against my normal CSV file (very similar to the uploaded test file).

With Python 2 I get this error: 

      (...)
      File "bank2ynab.py", line 41, in get_configs
        config.read(conf_files, encoding = "utf-8")
    TypeError: read() got an unexpected keyword argument 'encoding'

And with Python 3 I get this error: 

      (...)
      File "bank2ynab.py", line 89, in clean_data
        transaction_data = list(transaction_reader)
      File "/usr/lib/python3.5/codecs.py", line 321, in decode
        (result, consumed) = self._buffer_decode(data, self.errors, final)
    UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 541: invalid continuation byte

I then tried a search&replace for all the German umlaut characters in both lowercase&uppercase, but I still got the errors. I noted that the VISA CSV export has a `LF` line ending while the bank CSV export has `CR+LF` line endings. This is from the same bank, from the same online banking portal. I didn't try messing around with these line endings though.

I then used Notepad++ to convert the CSV file to ANSI file format. After that, the error in Python 2 remains but the script ran successfully in Python 3!

I can't make much out of this but I'm sure a detailed feedback might help you identify something I've missed.

Greetings from rainy Austria,
- Ben

nocalla:
After merging issue #29 into this, do you still get the same error here, @torbengb?

torbengb:
@nocalla I will test on my home Linux home tonight and respond here.

torbengb:
Test result with Python3: **success!** No errors found.

Test result with Python2:

      File "bank2ynab.py", line 197, in <module>
        main()
      File "bank2ynab.py", line 170, in main
        all_configs = get_configs()
      File "bank2ynab.py", line 34, in get_configs
        config.read(conf_files, encoding = "utf-8")
    TypeError: read() got an unexpected keyword argument 'encoding'

nocalla:
I wonder is that encoding parameter actually required now.

torbengb:
Well, if I leave out the `, encoding = "utf-8"` from line 34 then I get a different error:

      File "bank2ynab.py", line 196, in <module>
        main()
      File "bank2ynab.py", line 177, in main
        g_config = fix_conf_params(all_configs[section])
    AttributeError: ConfigParser instance has no attribute '__getitem__'

But it still works in Python3!

nocalla:
The python 2 version might be different to 3. May need to compare how they
handle things.

torbengb:
Hang on, I downloaded new CSV files and also downloaded a fresh copy of the bank2ynab.py in this pull request. 

Test result with Python3: **success!** No errors found.

Test result with Python2:

      File "bank2ynab.py", line 196, in <module>
        main()
      File "bank2ynab.py", line 170, in main
        all_configs = get_configs()
      File "bank2ynab.py", line 34, in get_configs
        config.read(conf_files, encoding = "utf-8")
    TypeError: read() got an unexpected keyword argument 'encoding'

Then I removed the `, encoding = "utf-8"`from the file and ran it again with Python 2: 

      File "bank2ynab.py", line 196, in <module>
        main()
      File "bank2ynab.py", line 177, in main
        g_config = fix_conf_params(all_configs[section])
    AttributeError: ConfigParser instance has no attribute '__getitem__'

Does this help you?

nocalla:
Need to do some reading about ConfigParser! But thanks.

toyg:
@torbengb 

> AttributeError: ConfigParser instance has no attribute '\_\_getitem\_\_'

I handled this exact error in the py2 branch, the ConfigParser interface is different in 2 and 3, so that's done.

> UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 

Eh, this is bitchy. Basically the encoding for these files is incompatible with utf-8 somewhere in its space. This is unsolvable with precision unless we include something like [`chardet`](https://pypi.python.org/pypi/chardet).

A possible strategy would be to do dummy `csv` runs through each file, surrounded by try/except, with the most popular charsets (utf-8, utf-16, iso-8859 variants etc) until one completes without error. 
```
def detect_encoding(filepath):
    for enc in ['utf-8',...]:
        try:
            with open(filepath, "rb", encoding=enc):
                for row in  csv.reader(transaction_file): continue
            return enc
        except ValueError:   # parent of UnicodeDecodeError and all that
           continue
```

This has obvious performance implications, but it might be acceptable in most cases - considering people will use this once a month or so, waiting a few seconds more should not be a biggie.

torbengb:
I like your idea of iterating through several charsets, it's actually an elegant solution. I hope it's worth the trouble! 

As for performance I'm sure it doesn't matter at all, given how the script already runs in the blink of an eye. I'd be surprised if it takes two seconds to complete this loop.

nocalla:
I quite like the try except way also. Interesting approach. I still feel it's a bit weird that I'm not getting that error on Windows though.

nocalla:
Made a pull request to merge @toyg 's branch into this so we can compare approaches. I think the py2 branch effectively quashes most errors but this one has most discussion attached! #54

nocalla:
Merged into the py2 branch. #54

PR #27, opened 2017-10-20
Cosmetics

torbengb:
@nocalla are you okay with the change in the config file?


torbengb:
I'll change the guide to `.bat` then. Thanks!

PR #36, opened 2017-10-21
Added UK Cooperative Bank format

toyg:


torbengb:
Thanks for this! Is there a reason why you didn't also add the file name pattern? I saw the other ticket about the UK sort code that would be meaningful to add (without the account number, obviously).

toyg:
Well, because it would be different for everyone... do you want me to add something suggesting values that should be changed, e.g.
```
Source Filename Pattern = yoursortcode_youraccountnum_
```

torbengb:
Oh, I thought the sort code would be the same for all customers of UK Cooperative Bank, but now I guess it's a separate code for each branch office then? Maybe the first two or three digits in the sort code are static for all branch offices? That's when wildcard matching could help us. If not even that, then I you did all you could.

@nocalla isn't the filename pattern a requirement for the script? Without that, the script won't be able to match any files against this section?

toyg:
There is a constant for the Coop sort codes, which is the starting "08". I guess I could add that.

nocalla:
@torbengb Yes. Without matching filename, nothing happens. 08 is too vague I would say. Workaround is to either manually rename export file or explicitly have user use format as @toyg specified above with Source Filename Pattern = yoursortcode_youraccountnum_. Wildcards are the actual solution.

nocalla:
Considering the feasibility of including a regex in the config for source filename. Thoughts?

torbengb:
I don't like regex because it looks like black magic voodoo to me (just like that ominous `vi` editor). But I recognize the power of regex and I think it would be a very useful addition.

nocalla:
Might as well merge this for now and add regex when it's sorted.

PR #39, opened 2017-10-22
Add python version check for csv writing

nocalla:
Issue #25 
Add python version check for csv writing
Also remove delimiter encoding as shouldn't be needed with unicode_literals and was causing me errors -
 wasn't working anyway according to @torbengb.
Also move `__future__` to top of file to prevent error.

```
SyntaxError: from __future__ imports must occur at the beginning of the file
TypeError: "delimiter" must be string, not bytes
```


torbengb:
Sorry I missed your request. I agree with the merge of course.

PR #43, opened 2017-10-22
Nicer cli output

torbengb:
- There's no need to list all the formats that aren't relevant to the user. Just state what format is used for each found file. 
- And having a "done" at the end makes it a lot easier to read when it processes several consecutive files.


nocalla:
It's good, except for one problem. 
`Specified directory not found, attempting to find Downloads folder.
Parsing file: elbi_umsaetze_.csv
Using format: AT Raiffeisen Bank
Writing file: fixed_elbi_umsaetze_.csv
Removing file: elbi_umsaetze_.csv
Done!
Specified directory not found, attempting to find Downloads folder.
Specified directory not found, attempting to find Downloads folder.
Specified directory not found, attempting to find Downloads folder.
Specified directory not found, attempting to find Downloads folder.
Specified directory not found, attempting to find Downloads folder.`

torbengb:
> except for one problem
I noticed that too - maybe the line 55-60 should be moved? Maybe `main()` should call a new `validate()` function where these things could live?

    b = g_config["input_filename"]
    ...
    if b is not "": 
        try:
            os.chdir(find_directory(g_config["path"]))
        except:
            print("Specified directory not found, attempting to find Downloads folder.")

nocalla:
I merged it as is, because it had a conflict with master that I resolved but maybe yes. I'd prefer validate_path myself, such is a bit more descriptive.

PR #44, opened 2017-10-22
Cosmetics

torbengb:
More little details, user guide, and helpful error messages.


torbengb:
I don't understand GIT enough to resolve this conflict. It has zero code impact though, so any resolution would work.

nocalla:
Looks okay to me now. It was just the comments. One line in the comments wasn't commented out and there was some weird repetition due to line number changes I think.

PR #45, opened 2017-10-22
summary of files processed and converted

torbengb:
This doesn't actually work :-(
The part about files_processed seems to work as it should, but the lines about files_converted complains about `UnboundLocalError: local variable 'g_files_converted' referenced before assignment`. Help?

### DIFF
--- a/bank2ynab.py
+++ b/bank2ynab.py
@@ -148,8 +148,7 @@ def find_directory(filepath):
     
 def main():
     # initialize variables for summary:
-    global g_files_processed
-    g_files_processed = 0
+    files_processed = 0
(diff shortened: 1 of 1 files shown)


nocalla:
Python doesn't like global variables. I'll look at it.

nocalla:
This should work okay now. Like I said in the commit, I don't think you need to treat "processed" and "converted" differently. If it matches the criteria for an account, we'll try to convert it!

torbengb:
Thank you so much for properly coding what I suggested! I appreciate your skills. And yes, you're right that we don't actually need both numbers.

PR #54, opened 2017-10-25
Merge our python 2 compatibility branches. 

nocalla:


nocalla:
#26

toyg:
The differences are basically:

* specifying "wb" when opening a file for writing in py2, which is recommended ["on platforms where that makes a difference"](https://docs.python.org/2.7/library/csv.html#csv.writer). That's a good point, but because of how much the branches now differ in the related code, it's hard to get it from issue-25. It would be much easier to make the change again (in _open_output) after merging py2 to master. 

* feature detection (`try/import/except`) vs version detection (`if version == 2`). [The strategy suggested in python docs](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection) is feature detection starting from most recent. It also avoids issues with alternative interpreters (jython, ironpython and so on) setting their own versions.

* the code in branch py2 sets a global switch that can be consistently reused anywhere we need.

So IMHO it would be easier to merge py2, then you can make the "wb" change directly on master.

torbengb:
I agree with @toyg's approach.

nocalla:
Sounds good to me! Is there any advantage in keeping the unicode_literals
import?

nocalla:
Also, why not use the try-except handling for our other version checks for config encoding and winreg import rather than if-else? Should we keep consistency across each?

toyg:
Ok, I did some more testing and there is still an outstanding issue with `csv` not working properly in py2 when source data contains non-ascii characters, regardless of how source or target files are opened and regardless of `unicode_literals`.

This is even mentioned [halfway through the examples](https://docs.python.org/2.7/library/csv.html#examples), providing a solution that is fugly but will probably work. I'll have a look online for alternatives.

This is why python 3 switched to all-unicode-all-the-time, py2 was hell for internationalization...

It would be good to set up some more formalized tests, btw.

torbengb:
@nocalla I opened issue #55 for your idea of formalized testing. 
I support the idea! But it will also cause overhead.

nocalla:
Think it wasn't my idea but I'm all for it!

torbengb:
corrected!

torbengb:
Any reason you didn't simply merge it right away? You don't need my permission for it :)

nocalla:
Just wanted to let it be looked over!

PR #70, opened 2017-10-28
Cosmetics

torbengb:
I want to pull the latest code into my branch and clean up my modifications. Then I'll merge it back into master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ A common project to consolidate all conversion efforts from various banks' expor
 
 ***Convert your downloaded bank statements into YNAB's input format.*** Here's what this script does, step by step:
 
-1. Look for and parse the `config.conf`. This file contains all the rules and import formats.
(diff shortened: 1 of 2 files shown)


nocalla:
You're actually merging into master with this pull, not from master into cosmetics as you state above.

torbengb:
Really?! Thanks for keeping me straight! I'm confused now because in the browser is says:

> Cosmetics #70
>  Open	torbengb  wants to merge 5 commits into master from Cosmetics

While I have your attention: a question -- I'd like to add a newline between each processed file, and at the start and end of the entire processing. I feel it makes the whole thing much more readable on the command line.  Any good arguments against that?

nocalla:
Exactly- into master.

Nope, doesn't bother me. I'll be running it from the batch file, so won't
even really see the output myself!

torbengb:
*I am an idiot.*

*I am an idiot.*

*I am an idiot.*

*I am an idiot.*

PR #71, opened 2017-10-28
Updating the Cosmetics branch, take 2

torbengb:


nocalla:
I always end up merging the branches when fixing conflicts. I must be doing something wrong. Anyway, you can merge this now and I've merged it into master by accident too I think!

PR #72, opened 2017-10-28
Cosmetic changes into master

torbengb:
Readme and CLI output lines, mostly.


torbengb:
Why are there no "files changed" in this? User error ID-10T.

PR #86, opened 2017-10-31
changed tests to read production conf file first,

toyg:
then add test.conf.


nocalla:
Question- what happens when there are 2 default sections like that? Which one overwrites?

toyg:
The last one to be read overwrites the previous one. I do it explicitly in the code so it's clear what is what; if you do it implicitly by just passing a list of files to read, I *guess* that it will follow the list order but tbh i never tried.

Issue #137, opened 2017-11-09
Support for N26 [DE] 


nocalla:
Name of Bank: N26
Country: Germany [DE]

Source: [DarkSmile92/YNABConverterPy](https://github.com/DarkSmile92/YNABConverterPy)

**To be filled in as required. Delete fields that match default values.**
Source Filename Pattern = example_transaction_export_filename
Source Filename Extension = .csv
Use Regex for Filename = False
Source CSV Delimiter = ,
Source Has Column Headers = True
Input Columns = Date,Payee,Outflow,Inflow,Running Balance
Plugin =

Master issue #21

diarmuidbourke:
I've made a patch to add support here: https://github.com/torbengb/bank2ynab/pull/149

nocalla:
Thanks for the input, @diarmuidbourke - as I mentioned on the pull request, the combined Inflow/Outflow is already accounted for, so there's no need to change the code for this bank - just the config file .

So, I've had a look at our survey and it looks like the config format should be as below. I've included the raw paste from the survey data too. I took the regex pattern from your pull request's code. Can you try this config with the main branch and see if it works?

```
[DE N26]
Source Filename Pattern = [0-9]{8}-[0-9]{8}
Input Columns = Date,Payee,skip,skip,Memo,skip,Inflow
```

Survey Data:
11/6/2017 15:18:17 | DE | N26 | Checking and MasterCard | Comma | Yes | Datum,Empfänger,Kontonummer,Transaktionstyp,Verwendungszweck,Kategorie,Betrag (EUR),Betrag (Fremdwährung),Fremdwährung,Wechselkurs | 2017-10-01,NINTENDO OF EUROPE,,MasterCard Zahlung,,Medien & Elektronik,-59.99,-59.99,EUR,1.0 | Date | Payee |   |   | Memo |   | Amount (EUR) (outflow: <0 inflow: >0) |   |   |   |   |   |   |   |   |   |   |   |   |   | robin@rkmail.email / https://github.com/DarkSmile92 | Reddit: DarkSmile92 |   | 20171001-20171106.csv |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

torbengb:
I believe I have completed this. Closing.

PR #149, opened 2017-12-31
#137 Add support for N26

diarmuidbourke:
N26 combines Outflow and Inflow into an Amount column.
This patch needs to split this column into the respective
Outflow or Inflow based on the sign of the number, negative being
Outflow, positive being Inflow.

### DIFF
--- a/bank2ynab.conf
+++ b/bank2ynab.conf
@@ -22,6 +22,7 @@ Input Columns = Date,Payee,Outflow,Inflow,Running Balance
 Output Columns = Date,Payee,Category,Memo,Outflow,Inflow
 Output Filename Prefix = fixed_
 Output Filename Extension = .csv
+Combined Outflow Inflow = False
 # Post-processing
 Delete Source File = True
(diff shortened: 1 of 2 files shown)


nocalla:
This patch isn't necessary, as this is actually already handled by the code, all you need to do is call your "Amount" column "Inflow" when deciding what the input columns are. YNAB handles importing a negative Inflow as being an Outflow.

If you look at the configuration file, you can see that there's another bank that works the same way. 

`[AT Raiffeisen Bank]
Source Filename Pattern = elbi_umsaetze_
Source CSV Delimiter = ;
Source Has Column Headers = False
Input Columns = Date,Payee,skip,Inflow,skip,skip,skip`

nocalla:
Closing this pull request as I've made a feature branch for N26 support over here: #150 
I've updated the config file there to reflect what I think will work for this bank. I'd be grateful if you could test out that branch to see if it works, @diarmuidbourke.

Thanks again.

Closed

PR #155, opened 2018-01-25
Feature/153 superfluous rows

### LINKED ISSUE #153: Add config parameters for header and footer lines to skip
@nocalla [had the idea](https://github.com/torbengb/bank2ynab/issues/140#issuecomment-359944921) to include a Start and End row feature in order to be able to ignore superfluous rows outside the main body of the input CSV.

As we're already skipping the first line when there is a header, perhaps we can re-work the `Source Has Column Headers` into something like `Skip header lines = 0` and `Skip footer lines = 0` where the values would be greater than zero if a header/footer is present.

We need to make sure that blank footer lines aren't counted, though. Blank header lines should probably be counted.

nocalla:
This fixes #153 .

Instead of having a boolean "Source Has Column Headers", we now specify "Header Rows = x" and "Footer Rows = y". 

I'm actually quite happy with how I implemented the change in code, but it could use some testing!

`output_data = output_data[header_rows:-footer_rows or None]`


nocalla:
I must fix the automated test files! Forgot!

nocalla:
The automated testing is a good reminder to do things like that!

torbengb:
Thank you Travis! :-)

nocalla:
It doesn't look like this feature breaks any existing banks - are you happy enough to merge?

PR #156, opened 2018-01-25
Pull feature/153 into feature/90?

torbengb:
I want to pull nocalla's new header/footer code into this Rabobank branch, let's see if I get it right :-)


nocalla:
Isn't this going against the whole Gitflow philosophy of working on features in isolation and merging only new stuff from Develop into the Feature branches? Not that there's huge differences, but I think you might end up confusing the merge into the final code from the feature branch.

nocalla:
I've merged the feature branch into develop, so I recommend you pull from there instead. Let's keep this nice and clean.

torbengb:
I agree with you - I guess I was getting ahead of myself in eagerness to try your stuff.

PR #162, opened 2018-02-01
Feature/161 inflow outflow flag

### LINKED ISSUE #161: Standard approach to using a credit-debit flag
[Issue 91](https://github.com/torbengb/bank2ynab/issues/91#issuecomment-362253461) discussed that the credit-debit flag is coming up often enough that it warrants incorporation or a more general plugin. 

The general issue here is that some banking formats don't have one inflow column and one outflow column, and they also don't have one amount column where outflows are represented using negative amounts. Instead, several banks use a format where one column contains an amount (always positive) and another column contains some value to indicate whether that amount is an inflow or outflow. 

This issue is created as a task to figure out a solution to this instead of having to create a format-specific plugin _just_ for that common functionality. (Until then, format-specific plugins are a viable solution.)

This ticket needs to deal with two aspects:

1. How should the "amount" column be referenced? -- We might just use the "inflow" here for simplicity, or invent a new "amount" parameter to clearly distinguish it from normal inflows.
1. How should the "c-d flag" column be specified? -- We would need a config parameter to turn on this c-d handling in the first place. We also need parameters to specify the actual values that indicate credit and debit (because some formats helpfully use C and D but others cryptically use A and B, and other values are sure to pop up later).

nocalla:
Fixes #161

### DIFF
--- a/bank2ynab.conf
+++ b/bank2ynab.conf
@@ -160,7 +160,9 @@ Input Columns = Date,Payee,Memo,Inflow,Outflow,skip
 [SE Swedbank]
 # Issue #134
 Use Regex For Filename = True
-Source Filename Pattern = exported_[0-9]{2}_[0-9]{2}_[0-9]{2}_[0-9]{4}
+Source Filename Pattern = exported
 Source Filename Extension = .csv
+Header Rows = 2
+Footer Rows = 3
 Input Columns = skip,skip,Date,Payee,Memo,Inflow,skip,CDFlag,skip,skip,skip,skip,skip,skip,skip
 Inflow or Outflow Indicator = 7,K,D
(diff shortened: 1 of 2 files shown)


nocalla:
In testing this, I've discovered that Rabo's date format does *not* get imported successfully into YNAB without the plugin.

torbengb:
I'm having a problem with Py2 but not with Py3. I'm putting it here and not as a bug because it's specific to this branch.

    $ python2 ./bank2ynab.py 
    
    Parsing input file:  /home/torben/Downloads/exported.csv
    Parsed 14 lines
    Writing output file: /home/torben/Downloads/fixed_exported.csv
      File "./bank2ynab.py", line 485, in <module>
        b2y.run()
      File "./bank2ynab.py", line 474, in run
        bank.write_data(original_file_path, output)
      File "./bank2ynab.py", line 428, in write_data
        writer.writerow(row)
      File "./bank2ynab.py", line 177, in writerow
        self.writer.writerow([s.encode("utf-8") for s in row])
    AttributeError: 'float' object has no attribute 'encode'
    $

This only happens for the `swedbank` format (I also tried my own `AT Raiffeisen`).  
This only happens with Python2. Python3 generates a correct output.

nocalla:
The Py2 error should be resolved by the last commit. Test when you can please, @torbengb.

torbengb:
Thank you! I am traveling now for a week, will be able to test again in ten
days.

nocalla:
Merging this, as it passes CI. We can run a bugfix if needed once you test it.

Issue #132, opened 2017-11-07
Support for Deutsche Kreditbank [DE]


nocalla:
Name of Bank: Deutsche Kreditbank [DE]
Country: Germany [DE]

Source:  [martinth/dkb-to-ynab](https://github.com/martinth/dkb-to-ynab) [Ruby]

**To be filled in as required. Delete fields that match default values.**
Source Filename Pattern = example_transaction_export_filename
Source Filename Extension = .csv
Use Regex for Filename = False
Source CSV Delimiter = ,
Source Has Column Headers = True
Input Columns = Date,Payee,Outflow,Inflow,Running Balance
Plugin =

Master issue #21

torbengb:
I have emailed the developer (Gmail) about the default source filename pattern. Let's see if he responds; his project is unchanged since 2+ years but people tend to react on email.

martinth:
Yep, they tend to :D

Admittedly, the project you linked never got beyond the proof-of-concept let-my-play-around-with-a-new-language phase (it's Rust btw, nor Ruby ;) ) but I think I can still help. 

The DKB actually has two separate output formats (that I know of). One for normal debit accounts and a separate one for credit cards. I will give you the details and also provide some (anonymized) sample data pulled fresh from my accounts.

# Debit accounts
Source Filename Pattern = Numeric account number. I think this is always 10 digits long but I'm not 100% certain.
Source Filename Extension = .csv
Source CSV Delimiter = ;
Source Has Column Headers = True (kind of, see below)
Input Columns = Payment date, Value Date, Type of booking, Payee, Description, Amount (negative for outflows), SEPA Creditor Identifier, SEPA Mandatsreferenz, SEPA Kundenreferenz

The format is a kind-of-CSV. It starts with a few lines that have metadata character and then there is the actual data. The metadata is also CSV (in the sense that it is fields, separated by delimiters) but its basically noise probably needs to be stripped, prior to parsing (but maybe it can be used to identify a DKB CSV export). Here is a sample:

```
"Kontonummer:";"DE42120300000000000000 / Girokonto";

"Von:";"25.12.2017";
"Bis:";"24.01.2018";
"Kontostand vom 24.01.2018:";" EUR";

"Buchungstag";"Wertstellung";"Buchungstext";"Auftraggeber / Begünstigter";"Verwendungszweck";"Kontonummer";"BLZ";"Betrag (EUR)";"Gläubiger-ID";"Mandatsreferenz";"Kundenreferenz";
[...]

PR #164, opened 2018-02-04
[DE Deutsche Kreditbank] Fix input reading

martinth:
For #132 I just checked a noticed two separate issues, that are both fixed here.

- 7ecaee2 fixes the credit card data import (and moves the description from the exported CSV to the "memo" section.
- ~~262d432 fixes a general issue with `Header Rows` in combination with `_valid_row` check, I _suspect_  that this is a general issue since the `_valid_row` check is applied before the `Header rows` offset. So `_valid_row` skips some lines and then `Header Rows` will discard some of the valid data~~


martinth:
@nocalla I removed the second commit and rebase onto `develop`.

nocalla:
Looks good. Thanks for your time in testing it!

PR #166, opened 2018-02-04
Feature/bugfix 165 row deletion conflict

### LINKED ISSUE #165: Bug: Header/Footer deletion occurs after check for valid rows, so number of rows is wrong!
Bug: Header/Footer deletion occurs after check for valid rows, so number of rows is wrong!
Solution: Apply header/footer deletion **before** checking for row validity. 

Thanks to @martinth for the heads-up in #164.

I'll have to have a look at this because I don't want to iterate through the entire file twice.

nocalla:
Fixes #165 

Skip header & footer rows when including rows in output. This occurs before row validation, preventing rows being deleted from messing up the header & footer count.

### DIFF
--- a/bank2ynab.py
+++ b/bank2ynab.py
@@ -169,6 +169,7 @@ class UnicodeReader:
     def line_num(self):
         return self.reader.line_num()
 
+
 class UnicodeWriter:
     def __init__(self, f, dialect=csv.excel, encoding="utf-8", **kwds):
         self.queue = cStringIO.StringIO()
--- a/bank2ynab.py
+++ b/bank2ynab.py
@@ -161,7 +161,7 @@ class UnicodeReader:
     def next(self):
         row = self.reader.next()
         return [unicode(s, "utf-8") for s in row]
-        
+
     def line_num(self):
         return self.reader.line_num()
 
--- a/bank2ynab.py
+++ b/bank2ynab.py
@@ -346,15 +346,20 @@ class B2YBank(object):
         cd_flags = self.config["cd_flags"]
         output_data = []
 
+        # get total number of rows in transaction file using a generator
+        with CrossversionCsvReader(file_path,
+                                   self._is_py2,
+                                   delimiter=delim) as row_count_reader:
+            row_count = sum(1 for row in row_count_reader)
+        
         with CrossversionCsvReader(file_path,
                                    self._is_py2,
                                    delimiter=delim) as transaction_reader:
             # make each row of our new transaction file
             for row in transaction_reader:
                 line = transaction_reader.line_num
-                print(line)
-                # skip header & footer rows [TODO FOOTER ROWS]
-                if line > header_rows:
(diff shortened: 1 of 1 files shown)


nocalla:
`AttributeError: UnicodeReader instance has no attribute 'line_num'`

...I hate Python 2.

nocalla:
```
python -m unittest discover
...Parsed 0 lines
FParsed 0 lines
F.
======================================================================
FAIL: test_read_data (test.test_B2YBank.TestB2YBank)
----------------------------------------------------------------------
  File "/home/travis/build/torbengb/bank2ynab/test/test_B2YBank.py", line 64, in test_read_data
    self.assertEqual(len(records), num_records)
AssertionError: 1 != 74
======================================================================
FAIL: test_write_data (test.test_B2YBank.TestB2YBank)
----------------------------------------------------------------------
  File "/home/travis/build/torbengb/bank2ynab/test/test_B2YBank.py", line 77, in test_write_data
    self.assertEqual(len(output_data), num_records)
AssertionError: 1 != 74
----------------------------------------------------------------------
Ran 6 tests in 0.024s
FAILED (failures=2)
```

I don't think my line_num attribute is working...

nocalla:
Okay, my grasping at straws didn't work, time to call for assistance.

@toyg - I need your help here please! How do you add a line_num attribute to your UnicodeReader class?

toyg:
having a look, 1 sec

toyg:
Ok, easy: csvreader.line_num is an attribute, not a method. Just remove those brackets and you should be fine.

nocalla:
Thanks! I think 2.7 is sufficient support and I'm regretting even agreeing to that!

torbengb:
Watching coders at work is a marvel, you guys are wizards!
As for Py2 how big of a problem would it actually be if we ditched 2.7 and required 3.x? It is wrong to assume that most people on Linux will understand, and most people on Windows don't have either installed already so they'd be indifferent too?

nocalla:
Heh. We'll see how things go.

toyg:
@torbengb i think it was you who first encountered the “py2 problem”, so it’s ironic you are calling for its removal ;)

Jokes aside, py2 is still the only python preinstalled on Mac, sadly, so personally I’d keep support around for a bit longer. It’s implemented here The Right Way™ by special-casing py2 rather than the other way around, so it will be trivial to rip out once that’s gone.

As an aside - sorry I’ve been away, it’s been a mess irl; I haven’t really used the script for a while (been a good boy and religiously entering all transactions as they happened...) plus you guys go at the speed of light, I can’t keep up :) but feel free to ping me when necessary and I’ll probably show up in a day or two max.

torbengb:
> it’s ironic you are calling for its removal

Yep! I just didn't know up front that even brand-new Linux releases only come with Py2 - tsk tsk! That's also why it's stated explicitly in our readme file. If keeping it is really no problem, then by all means, let's keep it!

PR #168, opened 2018-02-14
Date parsing

nocalla:
Use the strptime and strftime functions from the datetime library to parse date strings and return in a format YNAB can recognise.


nocalla:
Merging this, as it doesn't seem to have broken anything in the regular tests.

PR #169, opened 2018-02-16
Feature/#89 handelsbanken se

### LINKED ISSUE #89: Support for Handelsbanken [SE]
Hi, I got a request from @toyg to add this issue here since you were looking for more formats with input examples to add to this project.

This is about support for Handelsbanken, a Swedish bank.

The file extension is .xls, but the content of the file is HTML. See attached file for input example. Note that the actual data is anonimised, so the math does not add up.

[kontotransactionlist.txt](https://github.com/torbengb/bank2ynab/files/1435166/kontotransactionlist.txt)
(The attached file has extension .txt since it's not allowed to attach .xls files on github, but the real file should be .xls)

Since the format is unconventional you will probably need a plugin. I created a converter some time ago which may help: https://github.com/joacand/HandelsbankenYNABConverter

It's just a quick script and it probably needs be improved and changed to fit this project. You are free to use the code in any way you see fit.

**Edit from Nocalla (tentative config 15/02/2018)**
[SE Handelsbanken]
Source Filename Pattern = kontotransactionlist
Source Filename Extension = .xls
Input Columns = skip,Date,Payee,Inflow,skip
Source CSV Delimiter = ;
Header Rows = 7
Plugin = handelsbanken

From Issue #21 master list

nocalla:
Fix #89 using a plugin to strip out HTML formatting from the input file before our main read function acts on the file.

### DIFF
--- a/bank2ynab.conf
+++ b/bank2ynab.conf
@@ -155,7 +155,7 @@ Input Columns = Date,Payee,Memo,skip,skip,Inflow
 [SE Handelsbanken]
 Source Filename Pattern = kontotransactionlist
 Source Filename Extension = .xls
-Input Columns = Date,skip,skip,skip,Payee,skip,Inflow,skip,skip
+Input Columns = skip,skip,Date,skip,Payee,skip,Inflow,skip,skip
 Source CSV Delimiter = ,
 Header Rows = 6
 Plugin = handelsbanken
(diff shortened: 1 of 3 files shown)


torbengb:
That preprocess step is a sweet idea, wow!

Issue #50, opened 2017-10-23
Get our tool listed on YNAB website


torbengb:
Specifically at the end of these pages: 

- [YNAB 4 support page](https://classic.youneedabudget.com/support/article/csv-file-importing)
- [nYNAB help page](https://docs.youneedabudget.com/article/173-file-based-import)

nocalla:
Will they do that for a third party tool? Do they reference the toolkit
anywhere?

torbengb:
It wouldn't hurt to ask. The YNAB 4 page linked above lists several CSV tools, and they seem to be in various states of disrepair.

nocalla:
Worth a try. No point going there until we have the bugs out though.

On 23 Oct 2017 15:51, "Torben Gundtofte-Bruun" <notifications@github.com>
wrote:

It wouldn't hurt to ask. The YNAB 4 page linked above lists several CSV
tools, and they seem to be in various states of disrepair.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/torbengb/bank2ynab/issues/50#issuecomment-338684907>,
or mute the thread

nocalla:
Interestingly, there is a 3rd Party Github project listed on the [official YNAB site](https://docs.youneedabudget.com/article/921-importing-a-csv-file) now. 

It doesn't work all that well on our test data, but it's more user-friendly in that it's drag n' drop onto a [web page](https://aniav.github.io/ynab-csv/).

torbengb:
Interesting! So we would probably have a chance to get listed there too. I'll make another contact to YNAB.

I'm a bit confused and haven't tried out either of the two yet, but it seems that https://github.com/aniav/ynab-csv is forked from https://github.com/halloffame/ynab-csv and I wonder what the differences are.

Both of those tools have a step 3 saying *"For each column in the YNAB data file, choose which column you want to pull from your source data file"* and that sounds like the spot where we have a leg up?

nocalla:
Yeah, their one takes more effort per conversion, because you have to drag
things around a bit. Ours has a higher barrier of entry though, requiring
python installation and config file editing.

torbengb:
I'm submitting this message:

Subject:  
Website: link to free CSV conversion tool

Hi YNAB team!
I've seen that your website mentions CSV conversion tools here[1] and here[2]. Unfortunately, these require several steps for the user to do for every CSV file. 

[...]

PR #172, opened 2018-02-21
Update our "master" because it's really old

torbengb:
As I'm reviewing #50 and submitting a message to YNAB (the company), I realize that we should link to the master ZIP file but that one is 4+ months old and has only a few formats. 

This request is meant to update the master. Should we give it some kind of version number? Anything else we should take into account?


nocalla:
That's some changelog... I dunno are we ready to call it 1.0. But maybe we are, we have a working plugin system, support for several banks, and an automated testing system for newer code. 

I think we need to improve our tests so they verify that YNAB can import what's exported.

torbengb:
> are we ready to call it 1.0
Why not, and was that even a question? :-) 
More tests are always good, but I don't see it as blocking a release.

nocalla:
I'm okay with making this 1.0. We can make OFX output and better testing part of 1.1 maybe. The core functionality works.

bank2ynab.conf
```diff
@@ -30,38 +33,165 @@ Plugin =
...
+Footer Rows = 0
+Input Columns = Date,Payee,skip,skip,skip,CDFlag,Amount,skip,Memo
+Date Format = YYYYMMDD
+Inflow or Outflow Indicator = 5,Af,Bij
+
+[NL Rabobank]
+# survey response #14
+Source Filename Pattern = transactions
+Source Filename Extension = .txt
+Source Has Column Headers = False
```
nocalla: Old format, we need to fix this

bank2ynab.conf
```diff
@@ -30,38 +33,165 @@ Plugin =
...
+[SE Handelsbanken]
+Source Filename Pattern = kontotransactionlist
+Source Filename Extension = .xls
+Input Columns = skip,Date,Payee,Inflow,skip
+Source CSV Delimiter = ;
+Header Rows = 7
+Plugin = handelsbanken
+# don't delete source, so we can debug test
+Delete Source File = False
+
```
nocalla: Remove non-deletion

plugins/swedebank.py
```diff
@@ -12,15 +12,15 @@ def read_data(self, file_path):
         delim = self.config["input_delimiter"]
```
nocalla: Can we delete this plugin altogether?

torbengb: Yes; it's not referenced in the config.

bank2ynab.conf
torbengb: You're saying these lines should be removed, right? I'll do it now.

torbengb:
@nocalla re-review if you please?

nocalla:
We need to remove the test that specifically checks for the Swedbank plugin to make Travis happy!

torbengb:
> remove the Swedbank test

Is it just a matter of removing "bank2ynab/test/test_swedebankPlugin.py" too, or is there more to it than that? I don't really understand how it's made.

nocalla:
There's more to it then that I think. I'll tackle it soon, just got a lot
going on this week and next week.

torbengb:
My rule #1: "Don't stress." If you're busy with the RealLife(tm) then
please don't worry about this little project. It won't run away.

I'll delete the test file and see if that's enough.

torbengb:
Travis is happy now! Any reason to delay the merge @nocalla or can I go ahead?

nocalla:
Fire at will!

torbengb:
*BANG!*

PR #185, opened 2018-02-27
added Ostseesparkasse Rostock credit card & .gitignore

SebRut:
I added a template for credit card data for Ostseesparkasse Rostock (see attachments) and a .gitignore to prevent unused to files and directories like "__pychache__" or IDE dependent stuff to be committed.

[umsatz-1234________1234-20180227.zip](https://github.com/torbengb/bank2ynab/files/1761657/umsatz-1234________1234-20180227.zip)


bank2ynab.conf
```diff
@@ -121,6 +121,14 @@ Source CSV Delimiter = ;
 Input Columns = skip,skip,Date,skip,Memo,Payee,skip,skip,Inflow,skip,skip
 Date Format = %%d.%%m.%%y
 
+[DE Ostseesparkasse Rostock credit card]
+Use Regex for Filename = True
+Source Filename Pattern = umsatz-[0-9]{4}________[0-9]{4}-[0-9]{8}
```
torbengb: I'd prefer that 8-digit dates have the "20" explicitly, in order to prevent mismatches on 8-digit non-date numbers. So I'd make this end in `20[0-9]{6}` instead. Agree?

torbengb:
I've merged your changes into develop. I know it fails because develop itself is currently broken, but that is definitely not related to this merge.

Issue #184, opened 2018-02-26
Create more test data, and more test files


torbengb:
We can use the samples provided in [the survey](https://docs.google.com/spreadsheets/d/1KmDkLWEx-aVMlUWpNXafovt6B9GbeDDZ2OruaOp_aqE/edit?usp=drive_web&ouid=112556846578998968684) to create additional CSV files, including header row, data row, and proper file name.

Making an issue out of this so I don't forget.

torbengb:
I have added several files, but not integrated them into the test data scripts yet.

torbengb:
Closing as duplicate of #178.

PR #186, opened 2018-02-27
#184, proper regexes for dates and times.

torbengb:
I like the exactness of this approach, but I admit that the long regex strings look daunting. That's why I also added the samples in the comment section. I'm aware that [you disagreed here](https://github.com/torbengb/bank2ynab/issues/182#issuecomment-368596883) but I'm testing this anyway... 

I feel it would be careless to specify the patterns only as precisely as necessary, because when every new format changes the definition of "as necessary". We're already seeing some conflicts, and leaving the formats vaguely referenced opens us up to mismatches when future formats arrive. I think there's value in being precise, especially as the number of supported formats increase.

### DIFF
--- a/bank2ynab.conf
+++ b/bank2ynab.conf
@@ -35,7 +35,19 @@ Plugin =
 # Just make a new [section heading] and copy only the lines from above that need a non-default value.
 # Please keep this sorted alphabetically by [country + bank name] for easier maintenance.
 #####################################################################################################
-# DATE FORMATS:
+# FILENAME REGEX DATE FORMATS:
+# YYYY       ([12][0-9]{3})
+# mm         (0[1-9]|1[0-2])
+# dd         (0[1-9]|[12]\d|3[01])
+# YYYYmmdd   (([12][0-9]{3})(0[1-9]|1[0-2])(0[1-9]|[12]\d|3[01]))
+# YYYY-mm-dd (([12][0-9]{3})-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))
+# HH         (0[1-9]|1[0-9]|2[0-3])
+# MM         ([0-5][0-9])
+# SS         ([0-5][0-9])
+# HHMMSS     ((0[1-9]|1[0-9]|2[0-3])([0-5][0-9])([0-5][0-9]))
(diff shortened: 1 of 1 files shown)


bank2ynab.conf
```diff
@@ -35,7 +35,19 @@ Plugin =
 # Just make a new [section heading] and copy only the lines from above that need a non-default value.
 # Please keep this sorted alphabetically by [country + bank name] for easier maintenance.
 #####################################################################################################
-# DATE FORMATS:
+# FILENAME REGEX DATE FORMATS:
+# YYYY       ([12][0-9]{3})
```
torbengb: Is there a way to turn these samples into actual building blocks that we could use in the `Source Filename Pattern`? 

I.e. could we turn `([12][0-9]{3})` into e.g. `%%m` and so on? If we could, we could shorten:
`(([12][0-9]{3})-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))` to just:
`(%%y-%%m-%%d)`. 
This is more elegant and more readable - but we're inventing a new syntax and that may be a bad idea?

nocalla: I think that's an unnecessary amount of work. We'd have to introduce a whole dictionary of terms. Stick with simple regex. Why are we using these crazy long regexes anyway?

nocalla:
I don't see the point of this.

torbengb:
> I think that's an unnecessary amount of work. We'd have to introduce a
whole dictionary of terms. Stick with simple regex. Why are we using these
crazy long regexes anyway?

I sense that I am upsetting you with this, aren't I? That's not my
intention!

I think there's value in being precise, especially as the number of
supported formats increase. I feel it would be careless to specify the
patterns only as precisely as necessary, because when every new format
changes the definition of "as necessary". We're already seeing some
conflicts, and leaving the formats vaguely referenced opens us up to
mismatches when future formats arrive.

My goal was to prevent mismatches between dates and other "random" numbers.
If we specify that "this is an 8-digit date" then it won't match any other
8-digit numbers - or at least, the risk becomes so much lower.

torbengb:
I'll withdraw this for now because it sticks out like a sore thumb. I'll keep the branch around until we decide exactly how to handle conflicts (also that bug) - perhaps parts of this might be useful, the rest can go.

PR #193, opened 2018-03-03
update feature branch with fixed dev branch

torbengb:
I want to try how the fix works after packaging.


torbengb:
Nevermind this merge; I can do my test locally and not need to do this merge here.

Issue #126, opened 2017-11-04
Support for OCBC Bank [SG]


nocalla:
Name of Bank: OCBC Bank 
Country: Singapore [SG]

Source: [fongming/OCBC-YNAB-Parser](https://github.com/fongming/OCBC-YNAB-Parser) 

```
[SG OCBC Bank, checking and credit card]
# source: Issue #126
Use Regex for Filename = True
Source Filename Pattern = TransactionHistory_[0-9]{12}
```

Note: The source code ([direct link](https://github.com/fongming/OCBC-YNAB-Parser/blob/c3a5f9159a064d94799cfd9604213ff16af990d1/OCBC_YNAB_Parser.py#L40)) determines bank transactions or credit card based on the presence of certain words, and there seem to be different number of header rows depending on bank transactions ([here](https://github.com/fongming/OCBC-YNAB-Parser/blob/c3a5f9159a064d94799cfd9604213ff16af990d1/OCBC_YNAB_Parser.py#L12)) or credit card ([here](https://github.com/fongming/OCBC-YNAB-Parser/blob/c3a5f9159a064d94799cfd9604213ff16af990d1/OCBC_YNAB_Parser.py#L31)) - but apparently the same default filename.

Master issue #21

torbengb:
The source is based on Python, so perhaps @nocalla wouldn't have too much trouble grokking it?

The developer has been inactive for 2 years, so if you can't figure it out, let's close this issue as `wontfix`.

njunxuan:
Hi guys, I am interested in contributing on this issue as I do use this bank.

I have been looking at the code and based on what I gather, all that is necessary is for an appropriate conf to be created?

nocalla:
Hi! Thanks for the comment  

@torbengb is our documentation wizard, so he can reference the right docs, but it's very straightforward to make a config to be honest. 

What I'd suggest is that you paste an anonymised chunk of your sample data here. Or fill out the survey on the front page. 

Example config below:
To be filled in as required. Delete fields that match default values.
Source Filename Pattern = example_transaction_export_filename
Source Filename Extension = .csv
Use Regex for Filename = False
Source CSV Delimiter = ,
Source Has Column Headers = True
Input Columns = Date,Payee,Outflow,Inflow,Running Balance
Plugin =

njunxuan:
[...]

PR #200, opened 2018-04-18
Merge in njunxuan's fork.

nocalla:


torbengb:
I appreciate the contribution! But .... 

The test file is smaller than [the one in his own repo](https://github.com/njunxuan/bank2ynab/blob/9b90c2cdaeaab6ed92667f104107e8c0a306c0a9/test-data/TransactionHistory_20180418043121.csv), and more importantly it uses `XX.XX` instead of actual amounts. I'd like to see a better sample CSV file. 

As for the format config, and the "empty date value check", I think it's premature to merge this. We need to reach a consensus on #126 first.

Based on this, I think we should postpone this merge for a little while longer.

nocalla:
This is a merge into a test branch, not develop, don't worry.

nocalla:
I just wanted to have a start point for the plugin.

nocalla:
Also the test file in his own fork was just updated after I merged. I'll keep this branch up to date with his developments.

torbengb:
I stand corrected then :-)

PR #203, opened 2018-04-19
Hotfix/174 value error in python 2

### LINKED ISSUE #174: %%y-%%m-%%d causes "ValueError" in Python 2
Script language: Python 2.7
Operating system: MacOS X

What did you DO? (steps to reproduce)
[SE Nordea] using the date format `%%y-%%m-%%d` works with Python 3, but fails on Python 2.7.

What did you EXPECT to happen?
Should work with same format specification, regardless of Python version.

Can you provide other helpful information?
Described [here in Swedish](https://github.com/davols/nordea-ynab-convert/issues/1#issuecomment-367332527) (sorry!)  
For Python 2.7, the format needs to be `%y-%m-%d`.

nocalla:
Fix #174 

Trying out the hotfix function which uses the 1.0 version as a starting point, as that was the stable one.


bank2ynab.conf
```diff
@@ -177,7 +177,7 @@ Source CSV Delimiter = ,
 Header Rows = 1
 Footer Rows = 1
 Input Columns = Date,Payee,Outflow,Inflow,Running Balance
-Date Format = %%Y-%%m-%%d
+Date Format = %Y-%m-%d
```
torbengb: The 2 last configs use `MM/DD/YYYY` - should these also be changed to `%M/%D/%Y`?

nocalla:
I want to implement a test for date format before merging this.

bank2ynab.conf
nocalla: Yep. I just did a ctrl+F for %%!

nocalla:
Tests passed. Merging.

PR #205, opened 2018-04-29
Adjust to new file format of ING DiBa

jd1:
ING DiBa changed their file format:
* Filename: The files are now named `Umsatzanzeige_$IBAN_$Date.csv`.
* Header: There are now 13 rows of headers
* They swapped the columns _Saldo_(balance) and _Betrag_(amount)


nocalla:
Thanks for the update!

PR #213, opened 2018-05-23
Add Ceska Sporitelna (CZ) support

nejch:
Adds support for Ceska Sporitelna csv files.

For some reason they paginate output even for csv, so I added some test data for reference. But I haven't taken the time to check how the tests are run to include them.


nocalla:
Is the pagination just separate csv files? If so, there's no real need to handle it differently - the script will find all files matching the file pattern and convert them individually.

nejch:
Sorry I wasn't really clear there - yes, the pagination only affects the filename matching, hence the 3 different csv files included (no pagination, pagination in Czech and in English). I considered adding a test for the regex just as a quick sanity check (e.g. a `[test_cz_cs_regex]` section in test.conf). But I see now your test suite is focused mostly on the functionality of the core, not the actual bank configurations, so I won't pollute the test conf files for now.

BTW, have you considered using pytest instead of unittest as a runner to parametrize the tests? You have a lot of configuration and this way could generate separate tests based on the conf section (e.g. test_get_files[test_regex], test_get_files[test_num_files] etc.) and expose them explicitly in reports/CI instead of having to check the traceback to see what section failed. I can open a separate issue if you think it's worth it. [Here](https://docs.pytest.org/en/latest/parametrize.html)'s some documentation on this.

PR #220, opened 2018-08-03
Support for dk handelsbanken tested

duch11:
I tested for DK Handelsbanken, and it works seemlessly. 
I have changed README and Config to reflect this :)


nocalla:
It's great when one solution applies to many things! Nice work.

PR #223, opened 2018-10-11
BETA: Added support for sparekassen thy temp filename: thyposteringer

duch11:
Still need to know what the default filename is for sparekassen thy.
Don't know if that qualifies for a pullrequest however.. thanks

Issue reference: #219

Issue #219, opened 2018-07-31
Test Support for DK banks with online solution produced by "Bankernes EDB Central" (BEC)


duch11:
Hi while adding support for DK Arbejdernes Landsbank, I found out that many Danish banks uses the same online banking solution. It's produced by "Bankernes EDB Central" and therefore my update might have ended up adding many more Danish banks to the "Supported" List. 

Support added and verified for: 
- **Arbejdernes Landsbank DK** Verified
- **Merkur Andelskasse DK** Verified
- **Handelsbanken DK** Verified
- **Sparekassen Thy** Verified

The following list of banks, are potentially supported due to their membership with BEC, but I'm unsure if they all use same default settings etc in their export tool.

So, I need customers of these banks to test out if they work with my added config for BEC.

- Nykredit Bank
- Spar Nord
- VestjyskBANK
- Danske Andelskassers Bank
- Den Jyske Sparekasse
- GrønlandsBANKEN
- Fynske Bank
- Lollands Bank
- Salling Bank
- Lægernes Bank
- Totalbanken
- Møns Bank
- PenSam Bank
- Hvidbjerg Bank
- Frørup Andelskasse
- Andelskassen Fælleskassen
- Faster Andelskasse
- Frøslev-Mollerup Sparekasse
- Københavns Andelskasse

### Test with the following settings in the export tool:
**Date:** "whatever you need"
**Exportformat:** Default (CSV)
**Checkboxes:** "All unchecked"
**Filename:** Blank

How it should look:
![billede](https://user-images.githubusercontent.com/17031368/43408715-5192657a-9422-11e8-80fc-827592018758.png)

duch11:
**Handelsbanken DK** added in pull request #220

simzdk:
Mangler understøttelse for Sparekassen Thy: 
https://pastebin.com/cXMMkeHw

duch11:
Will be looking into it :)

duch11:
Hvad hed filen da du downloadede den? Det skal være hvad banken selv kalder filen.

nocalla:
I **think** it will include the header line in the final file, but I don't think it will crash or cause problems other than that.

duch11:
> 
> 
> Hej drenge,
> there's a checkbox `[ ] include column headings` that the user might check, so it would be prudent to be able to handle the CSV with and without a header row.
[...]

PR #227, opened 2018-10-12
Feature/sparekassen thy

nocalla:
Issue reference #219 

Merged from Pull Request #223

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,3 +103,6 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# visual studio code workspace
+*.code-workspace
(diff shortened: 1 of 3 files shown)


nocalla:
@duch11 @simzdk

So I don't think the error is down to quotation marks in the inflow/outflow. So here's what the script reads a row in the test data file for BOI as:

['01/09/2017', 'Random Name      GP', '', '428.03', '100']

By contrast, here's what it reads this test file's rows as:
['11-10-2018;"11-10-2018";"kontaktløs Dankort Føtex";"-248', '92";"600"']

For some reason, the entirety of the row is enclosed within single quotation marks, so it just sees a single entry in the row.

nocalla:
I had a hypothesis that the wrong bank format was being used to parse the file. Alas, this isn't the case.

nocalla:
Found the error - the delimiter was still a comma instead of a semi-colon! The config was never updated.

simzdk:
But the config I manually made I use the semi colon. How do I pull the develop branch to test?

nocalla:
That depends on your set-up, but there should be a download button on the front page if you just want to download it manually.

PR #245, opened 2018-11-22
Use env utility in shebang

eloyz:
Rather than have users specify which Python interpreter via the command line, we can allow the env utility to do this.

If `python3 ./bank2ynab.py` worked for the user, then `./bank2ynab.py` should also work for the user after this change.

[Made a GIF 😬](https://i.imgur.com/BA6duoF.gif)

### DIFF
--- a/bank2ynab.py
+++ b/bank2ynab.py
@@ -1,4 +1,4 @@
-#! /usr/bin/python3
+#!/usr/bin/env python3
 #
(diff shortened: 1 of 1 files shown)


eloyz:
Wanted to share this post as well. https://stackoverflow.com/a/7670338/2240803

nocalla:
Makes sense. Motion carried.

PR #246, opened 2018-11-22
Add support for US Bank of America

eloyz:
Adding support for [Bank of America](https://www.bankofamerica.com) CSV exports.

Bank of America aka (BOFA) doesn't have a separate inflow and outflow column.  Inflow and outflow is represented with a negative or positive number.

Ran unit-tests locally and everything is passing `python -m unittest discover -v`


nocalla:
Thanks for the submission! I must look into why Travis is failing though...

Issue #248, opened 2018-11-22
Travis is broken again


nocalla:
So something has changed in the last short while with Travis. It's reporting issues from flake8 regarding invalid escape sequences for strings that previously worked. 

Also, coverage and codecov installation is failing for everything pre 3.4.

nocalla:
I am at a complete loss here. Builds are failing very inconsistently across Python versions. 

We're getting the below output for every build that flake8 is tested in, but I'm not getting these results locally when testing. 

```
$ if [[ $TRAVIS_PYTHON_VERSION == 3* ]] && [[ $TRAVIS_PYTHON_VERSION != 3.2 ]]; then flake8 --builtins=unicode --ignore=W605; fi
./bank2ynab.py:93:37: E126 continuation line over-indented for hanging indent
./bank2ynab.py:98:29: E126 continuation line over-indented for hanging indent
./bank2ynab.py:225:13: E126 continuation line over-indented for hanging indent
./bank2ynab.py:506:17: E126 continuation line over-indented for hanging indent
./bank2ynab.py:528:18: W504 line break after binary operator
./bank2ynab.py:561:29: E126 continuation line over-indented for hanging indent
./test/test_B2YBank.py:44:17: E123 closing bracket does not match indentation of opening bracket's line
./test/test_B2YBank.py:62:17: E123 closing bracket does not match indentation of opening bracket's line
./test/test_B2YBank.py:120:29: E126 continuation line over-indented for hanging indent
./test/test_B2YBank.py:125:25: E123 closing bracket does not match indentation of opening bracket's line
./test/test_B2YBank.py:127:29: E126 continuation line over-indented for hanging indent
./test/test_B2YBank.py:131:25: E123 closing bracket does not match indentation of opening bracket's line
The command "if [[ $TRAVIS_PYTHON_VERSION == 3* ]] && [[ $TRAVIS_PYTHON_VERSION != 3.2 ]]; then flake8 --builtins=unicode --ignore=W605; fi" exited with 1.
```

Also, for every build prior to 3.4, the coverage tools are failing because they've both decided not to support these versions I *think*.
```
RuntimeError: Python 3.4 or later is required
The command "pip install codecov" exited with 2.
The command "pip install coverage" exited with 2.
```

nocalla:
So apparently this is down to an update in Flake8. Specifically an update in pycodestyle. 

2.4.0 (2018-04-10)
New checks:

```
[...]

PR #251, opened 2018-11-25
Updates to things in Travis checks

nocalla:
**New Pull Request**

**Reference Issue:**
#248 

**Description**
Travis is freaking out due to an update to flake8 and a minimum system requirement for the coverage checks.

- [x] pycodestyle update incorporated into Flake8. Escape sequences now checked more stringently
- [ ] coverage checking is requiring a minimum of Python 3.4

### DIFF
--- a/bank2ynab.py
+++ b/bank2ynab.py
@@ -273,8 +273,8 @@ def find_directory(filepath):
                 import winreg
             except ImportError:
                 import _winreg as winreg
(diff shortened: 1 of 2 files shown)


nocalla:
So I've temporarily disabled testing of Pyhton 2.7 because the detection of the Python version doesn't seem to be working in the test suite. I don't think the 2.7 version was being properly tested prior to me removing the conditional aspects of the .yaml file surrounding the format checking and coverage. I'll create a new issue for this for us to resolve.

PR #257, opened 2019-02-16
softer date parsing, combine multiple input columns, add support for [SK VUB]

milankowww:
**New Pull Request**
Support for [SK VUB] and required improvements to the infrastructure.

**Reference Issue:**
(none)

**Description**
I am submitting two improvements that are 100% backwards compatible with the existing configuration, but add more flexibility.

1. Dates with leading white space are now allowed. Previously, such dates could not be parsed and would result in exception.

1. Multiple input columns now can be combined to a single output column, by repeating the name of the output column in the *Input columns* configuration:
`Input Columns = Date,skip,Payee,skip,Payee,Inflow,skip,skip,skip,skip,skip,Memo,skip,skip,skip`

1. Both changes were necessary to add support for [SK VUB] which I am submitting as well. The bank in question uses two different Payee columns for bank transfers, and card payments.

Thank you for the useful project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,8 @@ Here is a list of the banks and their formats that we already support. Note that
 1. SE Swedbank
 1. SG OCBC Bank
 1. SG POSB savings
+1. SK Tatra Banka
+1. SK VUB
 1. UK Barclaycard credit card
 1. UK Co-operative Bank
 1. UK first direct checking
--- a/bank2ynab.conf
+++ b/bank2ynab.conf
@@ -476,3 +476,13 @@ Source CSV Delimiter = ,
 Input Columns = skip,Date,Inflow,skip,CDFlag,skip,skip,skip,Payee,skip,skip,skip,skip,skip,Memo
 Inflow or Outflow Indicator = 4,Kredit,Debet
 Date Format = %d.%m.%Y
+
+[SK VUB]
+Source Filename Pattern = SK030200[0-9]{16}_export_(of_movements_on_the_account|pohybov_na_ucte)
+Source Filename Extension = .CSV
+Use Regex For Filename = True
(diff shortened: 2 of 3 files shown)


torbengb:
Re: your item no 2, how are the two fields combined? What happens if there
is data in both of them? Are they separated by a space character?

milankowww:
> Re: your item no 2, how are the two fields combined? What happens if there is data in both of them? Are they separated by a space character?

yes, that is right:
`cell = " ".join(cell_parts)`

torbengb:
2: sounds great!

1: given no 2, it might be prudent to trim whitespace from every field, not
just dates. If not for this bank, then for who knows what future bank.

milankowww:
> 2: sounds great! 1: given no 2, it might be prudent to trim whitespace from every field, not just dates. If not for this bank, then for who knows what future bank.

Some fields might contain leading spaces already, so doing this globally may change how the Payee and Memo looks like in the new version. I still think it's a good idea though.

I have made a new commit to the open pull request. It strips leading whitespace for all fields, and also makes the CI checker happier - I noticed the unit test fails when there is incorrect number of input fields, so I have reintroduced the IndexError exception handling, also for Date which would fail even in the original code.

milankowww:
done; I've cleaned up the automated checks as much as possible:
- travis CI and flake8 are now happy (had to clean up some original code as well);
- codeclimate is still nervous about "cognitive complexity" of the new code. It probably fears lambda, but it cannot be helped - the current version is much more elegant than when rewritten without "filter" which is using the lambda.
- code coverage has only decreased because the test cases don't cover the pathologies

nocalla:
And yes, the test cases are incomplete as it stands, unfortunately.

PR #262, opened 2019-03-31
WIP: Feature/183 ynab api

nocalla:
**New Pull Request**

**Reference Issue:**
#183 YNAB API Access

**Description**

In-progress! Pull request to track development of branch only.

Key Changes:
- Add API functionality
- Add API token and API account sections to config
- user_configuration.conf is used to store API and personal settings - make sure not to accidentally push this file, devs!
- Several steps required to standardise output for API requirements
    - Standardise decimal points in output
    - convert date to YYYY-MM-DD
- Attempt to add date format to each bank format
- Create requirements.txt and Requests requirement
- Lots of unit testing for API functions


duch11:
Hello there! 
Lovely work, it sparked some questions in my mind of how we want to do things, I've posted them in the issue. #183

torbengb:
It doesn't happen automatically. On manual import there's a checkbox to
swap memo and payees, but that's not available in the API of course.

I prefer to define this in the format config because that is more
transparent than doing it in the code. The user might be expected to look
at the config but he should definitely not read the code.

We could introduce a format parameter to mirror the checkbox, but I think
that is needless extra effort as it only does the same as changing the
column mapping in the same format.

torbengb:
How about a format parameter to turn on a bit of code that COPIES the memo
into payee? That would be my favorite result in the converted file because
I would retain the comment while also taking advantage of YNAB's payee
management.

nocalla:
No, you're right actually - it's better to just name it in the config as Payee. That way our auto_memo function will just use it in both places - preserving the original memo and providing a payee that can be remapped by YNAB.

torbengb:
Ah, does that auto_memo also apply to the API? It should...

nocalla:
Yeah, the API just processes the output after we've done all our fixes to the format.

nocalla:
Functionality is working. Needs refinement, but for now I'm going to switch focus to improving code coverage. I'd like to make tests for every function in the API class, so this may take some time and may result in rewriting of some functions.

duch11:
Nocalla! I must say I'm impressed with your progress! 
You are a true beast for this project! 
Thank you so much for your hard work! 

I feel so bad I don't have time for contributing much ATM, but right now I'm doing my finals and need to focus on that. 

But from July and forward I'll be looking to contribute a heck of a lot more!

nocalla:
Don't worry - focus on studying and ignore this! I use this project to learn about programming - I'm experimenting with testing at the moment, so I'm entertained.

nocalla:
As I'm under a lot of time pressures currently, I haven't been doing much work on this. What do you think, @torbengb  - should I push this to `develop`? It seems to work well and I think the only way we're going to see more issues is if a wider audience is aware of the API functions.

torbengb:
Good call, I agree.
I too have more than enough on my list, so apologies for not getting back
to you sooner.

We still have a handful of edge cases (identical filename, different
format) but that should not stop this development. I use it a lot myself,
works fine! Thinking about a new public release...

PR #272, opened 2019-04-17
Add ANZ/NAB/ING Australian banks

jdgordon:
Adds support for 3 Australian bank exports.

* ANZ (via ANZ.csv)
* ING (via Transactions.csv)
* NAB (via TransactionHistory.csv)

### DIFF
--- a/bank2ynab.conf
+++ b/bank2ynab.conf
@@ -82,6 +82,31 @@ Source CSV Delimiter = ;
 Header Rows = 0
 Input Columns = skip,Date,Memo,skip,skip,skip,Inflow
 
+[AU ANZ]
+# Transactions.csv
+Source Filename Pattern = ANZ
+Header Rows = 0
+Input Columns = Date,Inflow,Memo
+Date Format = %d/%m/%y
+
+[AU ING]
+# Transactions.csv
+Source Filename Pattern = Transactions
+Header Rows = 1
+# Double 'Inflow' looks strange but is correct
+# The input csv uses 2 seperate columns, but has positive values
+# for the inflows and negative values for the outflows
+# The conversion will take those negative inflows and convert them to outflows
+Input Columns = Date,skip,Memo,Inflow,Inflow
+Date Format = %d/%m/%y
+
+[AU National Australia Bank]
+# TransactionExport.csv
+Source Filename Pattern = TransactionHistory
+Header Rows = 0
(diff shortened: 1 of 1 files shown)


jdgordon:
@nocalla 

> Thanks for the contribution. It might make more sense to use Payee instead of Memo in these fields. Memo is automatically filled in with the Payee details by the script and then Payee details can be used by YNAB directly for renaming purposes.
> 

"V2718 16/04 GLICKS CAKES/BAGELS PL BALACLAVA " is an example of one of those memo lines, they have way more information (and are mostly not static) that I'd expect to put in the payee field, happy for you to change it to payeeif you feel thats more inline with the other importers though.



> Does the double Inflow column work properly if there's numbers in both columns of the same row?

I cant imagine ever having both values set, and presumably the import code would just take the second field. (cant test now unfortunately)

bank2ynab.conf
```diff
@@ -82,6 +82,31 @@ Source CSV Delimiter = ;
 Header Rows = 0
 Input Columns = skip,Date,Memo,skip,skip,skip,Inflow
 
+[AU ANZ]
+# Transactions.csv
+Source Filename Pattern = ANZ
+Header Rows = 0
+Input Columns = Date,Inflow,Memo
+Date Format = %d/%m/%y
```
torbengb: Just checking: do these formats really have 2-digit years? 
(4-digit would be uppercase %Y)

nocalla:
I think that would be more in line with the functionality of the other formats, as it allows for automatic conversion by YNAB of a Payee name containing, say, GLICKS CAKES, to a payee called Glicks Bakery. So the extra info will be culled if it's a repeat Payee or can just be ignored and deleted. The script relies on the presence of a Payee field to some extent and this method preserves the info in the Memo field without losing any info. 

The logic of the multiple input columns is to accommodate multiple text fields, e.g. memos spread across a few cells. The values all get merged into one, which I don't think would work well with numbers! What's the logic of the 2 inflow columns? Is it at all possible that the two could get filled together? If not, then that's fine.

torbengb:
Regarding memo-vs-payee, I have learned that YNAB's payee management is very powerful. My own bank doesn't provide payee information but the memo can be used. 

... aaand I saw that Nocalla has posted exactly this while I am typing, so I'll just stop right here :-)

nocalla:
Merging this. We can amend the year if we get more feedback.

PR #280, opened 2019-05-31
Correct "BE KBC Checking" Input Columns

dotEsuS:
**New Pull Request**

**Description**

- Add Payee column header.
- Modify Memo column header. 
The previous Memo column is a redundant copy of multiple columns, but the real Transaction Memo is the last column.


torbengb:
This looks good - thank you very much for the contribution!

PR #284, opened 2019-07-28
Fix/lansforsakringar

diblaze:
The Regex for the Swedish bank "Länsförsäkringar" is not correct. As each account can have a different name, it is instead better to use the ".+" qualifier than a statically typed account name.

The bank also used "Meddelande" (memo) as it's Payee for some reason. Therefore I also changed this to correctly correspond to the banks intention. This can be changed manually in YNAB by ticking the "Switch Payee and Memo" button, but why not automate it?

### DIFF
--- a/bank2ynab.conf
+++ b/bank2ynab.conf
@@ -466,7 +466,7 @@ Plugin = handelsbanken
 # source: survey response #43
 # Mitt konto senaste transaktioner 2019-05-28.csv
 Use Regex for Filename = True
-Source Filename Pattern = Mitt konto senaste transaktioner [0-9]{4}-[0-9]{2}-[0-9]{2}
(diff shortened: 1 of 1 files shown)


diblaze:
Also would be ideal to somehow remove the second line of the .CSV after generating the new .CSV file. As the second line has nothing to do with the transactions.
This is the second line I am talking about:
`Date,Payee,Category,Memo,Outflow,Inflow
Privatkonto,"35 609,55",,"35 609,55",,"35 609,55"`
It only shows the amount of money available on the account, and AFAIK it is useless for YNAB importing?

torbengb:
Hello Denis!
I'd like to follow up on this issue. Did we fulfill your wish to remove line 2 of the CSV? If not, I think we should be able to address that quite easily! ("Header Rows = 2")

Issue #291, opened 2019-10-13
Re-enable Deutsche Kreditbank checking account in config again


lukkigi:
**Is your feature request related to a problem? Please describe.**
The template for DKB checking accounts is still commented out, even though #182 has been solved.

**Describe the solution you'd like**
Re-enable the template section again. I have tested a DKB checking account and it seemed to work fine.

**Additional context**
I will be creating a PR for this issue in the next days

PR #292, opened 2019-10-13
Re-enable DE Deutsche Kreditbank checking in config

lukkigi:
**New Pull Request**

**Reference Issue:**
*#291*

**Description**
The Deutsche Kreditbank checking account section should be enabled in the config again

I have removed the comments from the Deutsche Kreditbank checking account section and thus enabled it again


nocalla:
Makes sense, we should have done this when merging the fix for #182.

Issue #293, opened 2019-10-13
Transactions with amounts bigger than 999€ get imported without a value


lukkigi:
**Describe the bug**
Transactions with amounts bigger than 999€ get imported without a value

**What did you EXPECT to happen?**
I wanted to import all my transactions to YNAB.

**What ACTUALLY happened?**
Any transaction with amounts like 1500€ or 5000€ were imported without any value.

**What did you DO? (steps to reproduce)**
1. Download the csv
2. Run bank2ynab.py

**What's your software environment?**
 - Script language: Python 3.7
 - Operating system: Linux
 - OS version: Arch Linux

**Can you provide other helpful information?**
Values in the original csv file were formatted like this: 1.200,00€
Values in the fixed csv file were formatted like this: 1.200.00€

I will be opening a PR for this issue.

PR #294, opened 2019-10-13
Add support for german number formatting

lukkigi:
**New Pull Request**

**Reference Issue:**
*#293*

**Description**
German numbers formatted like 1.200,00 are now supported by a new config option called "Number Format".

Due to German number formatting, amounts were outputted like this: 1.200.00, thus failing YNAB import. I have added a new config option "Number Format" currently supporting "de" and just empty. If it is "de", the numbers will be formatted correctly now. I have also edited the default config sections for DKB checking and DKB credit card accounts (the only German banks I was able to verify this issue with).

### DIFF
--- a/bank2ynab.conf
+++ b/bank2ynab.conf
@@ -15,6 +15,8 @@ Footer Rows = 0
 Input Columns = Date,Payee,Outflow,Inflow,Running Balance
 # (see https://docs.python.org/2/library/datetime.html#id1 for date format strings)
 Date Format =
+# Currently either de or empty for the default number formatting
+Number Format =
 Inflow or Outflow Indicator =
 # Output file formatting
 Output Columns = Date,Payee,Category,Memo,Outflow,Inflow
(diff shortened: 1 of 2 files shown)


nocalla:
I agree that this is a necessary change, but I'm wondering about the implementation. I want to review this before merging and maybe make it more multi-purpose for potential use with other number formats.

nocalla:
I think I have a better option than requiring a new configuration parameter. Once the fix_decimal_point function is finished its conversion, we just remove every instance of "." except for the last. That should allow for different arbitrary number separators in future if they occur in different formats and avoids any additional functions.

New fix_decimal_point code for reference, please test and see if it works in your case, @lukkigi. It works with the simple test data I made for it, but might throw up something with real-world data. I've created a pull request over at # if you want to try the build.

```
def _fix_decimal_point(self, row):
        """
        convert , to . in inflow and outflow strings
        then remove every instance of . except last one
        :param row: list of values
        """
        inflow_index = self.config["output_columns"].index("Inflow")
        outflow_index = self.config["output_columns"].index("Outflow")        
        inflow = row[inflow_index].replace(",", ".")
        outflow = row[outflow_index].replace(",", ".")        
        dot_count = inflow.count(".") - 1
        row[inflow_index] = inflow.replace(".", "", dot_count)
        dot_count = outflow.count(".") - 1
        row[outflow_index] = outflow.replace(".", "", dot_count)

        return row
```

PR #295, opened 2019-10-13
Updated the README to mention the automatic import capabilities

lukkigi:
**New Pull Request**

**Reference Issue:**
*I wasn't sure if I need to create an issue for a simple README change. If this is needed, I will create one.*

**Description**
The README did not mention the automatic import and I was quite surprised about that when it works so flawlessly and is definitely a huge feature of the project.

I have added a section to the README mentioning the automatic import and the need for the YNAB API access token.


torbengb:
Thank you for this! You're right that this is an awesome and **major** feature in our tool, and it has definitely earned a mention in the readme file. 

Having said that - you might have over-promised a little in your contribution :-) because it's not quite polished yet. It will remember the format-to-account mapping but it will ask you every time if you have more than one budget. (And we haven't yet solved the case when you have more than one account with the same format.)

Despite this caveat, I'll go ahead and merge your contribution (lightly edited). Thank you!

PR #297, opened 2019-10-27
Feature to set date format language

blinkiz:
If source date format is like '16 Oktober 2019' and 'Oktober' is spelled in another language then your current LC_TIME, then you need to change date format language.
Valid inputs are 'sv_SE.UTF-8', 'en_US.UTF-8' or whatever your operating system understand as LC_TIME variable.

### DIFF
--- a/bank2ynab.conf
+++ b/bank2ynab.conf
@@ -15,6 +15,7 @@ Footer Rows = 0
 Input Columns = Date,Payee,Outflow,Inflow,Running Balance
 # (see https://docs.python.org/2/library/datetime.html#id1 for date format strings)
 Date Format =
+Date Language =
 Inflow or Outflow Indicator =
 # Output file formatting
 Output Columns = Date,Payee,Category,Memo,Outflow,Inflow
--- a/bank2ynab.py
+++ b/bank2ynab.py
@@ -26,6 +26,7 @@ from datetime import datetime
 import logging
 # API testing stuff
 import requests
+import locale
 import json
 
 # configure our logger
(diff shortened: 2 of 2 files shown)


nocalla:
Thanks for the contribution. From what reading I've done of this in the past though, I think there's some issues with locale and different operating systems. I'm not familiar with it myself but I think this change may need some testing before we merge. Are you familiar with the issues I've mentioned?

blinkiz:
> Thanks for the contribution. From what reading I've done of this in the past though, I think there's some issues with locale and different operating systems. I'm not familiar with it myself but I think this change may need some testing before we merge. Are you familiar with the issues I've mentioned?

Yes. I also notice people saying that LC_TIME string is different between Linux and Windows. Another solution would be to use a python package that can handle languages. What I have found is babel. 'from babel.dates import format_date, format_datetime, format_time'.

The reason for creating this pull request is because the bank Revolut uses the string '16 Okt. 2019' and currently bank2ynab do not understand "Okt" with my LC_TIME language settings.

nocalla:
I think this would this be more suitable to be implemented as a plugin. We could call the plugin something bank-neutral so it could be used for other banks that have similar dates. 

Then we aren't relying on a function that is unpredictable between operating systems in a way that may end up affecting all banks.

nocalla:
@blinkiz - I'm going to put together a plugin for this. Can you please provide a sample few lines of the input file so we can test the format?

blinkiz:
Hello. Sure, I can provide lines to process. Thank you!
I will reply back on Monday with csv lines.

blinkiz:
Hello. I have replied in #300. Have a good day!

PR #311, opened 2020-01-03
Feature debug and api fix

### LINKED ISSUE #231: Logging levels & configuration
*Dear submitter, please edit the following according to your needs. Thank you!*

**Other feature request**

At the moment, when running the automated tests, all of the messages spam the test console. 

Implement logging levels and a switch to disable the not-required ones when testing. 
This will also have the benefit of allowing us to add additional debug lines for hunting errors if we need them.

Otherwise we could just remove **all** of the output messages! 

**Additional context**
Some references about logging:
- [StackOverflow Question](https://stackoverflow.com/questions/6918493/in-python-why-use-logging-instead-of-print)
- [5 minute quickstart guide to python logging module](https://inventwithpython.com/blog/2012/04/06/stop-using-print-for-debugging-a-5-minute-quickstart-guide-to-pythons-logging-module/)
- [Gist with practical implemenations of logging](https://gist.github.com/neuroticnerd/7c60d61c8d9d9716f50d)

@toyg's take on this issue from #214:
`Step 1: the logging level must be set in a config file, not in source. The simplest way to achieve this is to have a config parameter that can store a value (e.g. LogLevel = INFO), then in code something like:basicConfig(... , level=getattr(logging,your_param)) (you might want to have a check for acceptable values before: if your_param not in ['INFO','DEBUG'...]: your_param='INFO')Step 2: if you want to do something more complicated/precise, have a look at dictConfig().Step 3: you can have a simple environment variable (say DEBUG=1) and depending on that value, switch between basicConfig and dictConfig, so you can e.g. pass a complicated config for debug/testing purposes, and a simple one for production default. Or just keep all the logging machinery in a separate config file (like a JSON) and use dictConfig all the time, providing different files when required.I'm not a test guru, I'm sure there are many other ways to skin this particular cat, but the ideas above are the simplest strategies and rely on simple stdlib modules.`

clancychilds:
**New Pull Request**

**Reference Issue:**
Fixes #231 
Fixes the inability to work with the API due to date formatting issues. Also adds a DEBUG logging level for easier debugging.

Adds a bit of code to change the date format that is provided for the file uploads (mm/dd/yyyy) to the format needed for the API (yyyy-mm-ddd). Also adds the ability to set a `Log Level = DEBUG` for more descriptive debugs
As well: provides more detail in the output about why an API request has failed.

### DIFF
--- a/bank2ynab.py
+++ b/bank2ynab.py
@@ -825,7 +825,7 @@ class YNAB_API(object):  # in progress (2)
                    self.budget_id,
                    self.api_token))
         logging.debug(f"URL: {url}")
-        logging.debug(f"Payload: {data}")
+        logging.debug(f"Payload: {json.dumps(data)}")
         post_response = requests.post(url, json=data)
 
         # response handling - TODO: make this more thorough!
@@ -909,8 +909,9 @@ class YNAB_API(object):  # in progress (2)
         }
         id = details["id"]
         name = details["name"]
-        detail = errors[id]
-        logging.error("{} - {} ({})".format(id, detail, name))
+        error_reason = errors[id]
+        detail = details.get("detail", "No further detail provided")
+        logging.error("{} - {} ({}) - {}".format(id, error_reason, name, detail))
 
         return ["ERROR", id, detail]
 
--- a/bank2ynab.py
+++ b/bank2ynab.py
@@ -778,12 +778,12 @@ class YNAB_API(object):  # in progress (2)
         date_matcher = re.compile(date_pattern)
         date_match = date_matcher.fullmatch(date)
         if (date_match):
-            logging.debug(f"Reformatting date: {date}")
+            logging.debug("Reformatting date: {}" % date)
             yyyy = date_match.group('yyyy')
             mm = date_match.group('mm')
             dd = date_match.group('dd')
-            date = f"{yyyy}-{mm}-{dd}"
-            logging.debug(f"Reformatted date: {date}")
+            date = "{}-{}-{}" % (yyyy, mm, dd)
+            logging.debug("Reformatted date: {}" % date)
         else:
             raise(ValueError)
         payee = this_trans[1]
@@ -837,7 +837,7 @@ class YNAB_API(object):  # in progress (2)
                "{}/transactions?access_token={}".format(
                    self.budget_id,
                    self.api_token))
-        logging.debug(f"Transaction Payload: {json.dumps(data)}")
+        logging.debug("Transaction Payload: {}" % (json.dumps(data)))
         post_response = requests.post(url, json=data)
 
         # response handling - TODO: make this more thorough!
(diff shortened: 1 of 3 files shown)


nocalla:
Hi @clancychilds thanks for the contribution. I very much appreciate the logging input! The date fix shouldn't be necessary though, as output date is automatically converted to the YNAB API form. I haven't reviewed your changes in detail, but I'm wondering is it the bank format in the conf file that's causing the problem you're having.

nocalla:
`# fix the date format
row = self._fix_date(row, date_format)`

I knew this was in there. If "date format" isn't specified in the config file for the bank, it won't work unfortunately. So if you review your bank's config and add the date format, the date fix isn't needed.

clancychilds:
I'm a bit confused by the expected date formats, @nocalla : 
In https://github.com/bank2ynab/bank2ynab/blob/45e6ad9824c4968ca30f116373e5b52c7352684b/plugins/plugin-template.py#L30 it mentions  `YNAB's date format is "DD/MM/YYYY".` And that is what I followed when I built out plugins/JLP_Card_UK.py

But the _fix_date function you referenced seems to output in yyyy-mm-dd format.

Which format should my plugin be outputting for the date of a transaction? Will it still work for both file uploads and API integration?

nocalla:
Ah, so the comment in the plugin wasn't updated when we pushed the API functionality. I'll update that. 

Ideally, a plugin should use the `_preprocess_file` function to provide the data upstream of the `read_data` function. That way, if there are any changes to `read_data`, the plugin doesn't need to be maintained. 

If you're reproducing the `read_data` function, output dates in a YYYY-MM-DD and it should work fine for now.

I'm parsing your plugin to try to establish why a plugin is necessary for your bank.

- Date format is "DD-MON-YYYY" - the config file accepts standard Python date format strings for conversion by `_fix_date`
- CR means inflow - `_cd_flag_process` is for converting this, though I've just coded it to search for an outflow indicator - is there a corresponding debit flag?
- Pound sign regex - I can't work out what this is doing, is it just removing £ from all of them?

So in theory, all your plugin needs to do is to do whatever you're doing with the pound sign. And I'm actually not even certain if the API pays attention to currency signs! Can you let me know what the regex is doing?

bank2ynab.py
```diff
@@ -766,6 +773,21 @@ def process_transactions(self, transaction_data):
 
     def create_transaction(self, account_id, this_trans, transactions):
         date = this_trans[0]
+        # API requires yyyy-mm-dd format rather than
```
nocalla: Please remove date section for handling in plugin, as per previous comment.

nocalla:
Merging this into a branch for the logging feature.

clancychilds:
I'm a bit lost as to what just happened here. If I want to continue making the fixes required for my plugin, do I need to rebase my working copy on that new branch you created?

nocalla:
Sorry, I was a bit in the zone today so was moving fast. I liked your work on the logging config, so made that its own branch for further tweaking. (Which I didn't do in the end because I got distracted by GitHub Actions.) If you want to add more to that, that's the branch to target. 

Your plugin work shouldn't require anything from the logging branch, so there's no need to rebase based on that. I'd suggest basing it off develop, as I've updated the plugin template there to a more useful example of the preprocess function.

PR #313, opened 2020-01-07
Logging Configuration (231)

### LINKED ISSUE #231: Logging levels & configuration

nocalla:
**New Pull Request**

**Reference Issue:**
Fixes #231 

**Description**
Allows configuration of logging for debug usage.

Incorporates changes from @clancychilds


bank2ynab.py
```diff
@@ -646,6 +646,13 @@ def __init__(self, config_object, is_py2=False):
...
         self.transaction_data = {}
 
+        # switch logging level if defined (the earlier the better)
+        log_level = config_object.get('DEFAULT', 'Log Level').upper()
+        numeric_log_level = getattr(logging,
+                                    log_level.upper(), None)
+        if not isinstance(numeric_log_level, int):
+            raise ValueError('Invalid log level: %s' % log_level)
+        logging.getLogger().setLevel(numeric_log_level)
         for section in config_object.sections():
```
nocalla: Fix quotes and string format to match rest of file

bank2ynab.py
```diff
@@ -766,6 +773,21 @@ def process_transactions(self, transaction_data):
 
     def create_transaction(self, account_id, this_trans, transactions):
         date = this_trans[0]
+        # API requires yyyy-mm-dd format rather than
```
nocalla: Remove all changes to date logic

nocalla: Leave check in just in case, but remove reformatting step.

bank2ynab.py
nocalla: Resolved

nocalla:
Stale branch - features already incorporated.

PR #316, opened 2020-01-09
Feature/275 remember format budget link

### LINKED ISSUE #275: Remember format/budget link
**Current**

The new sweet API integration cleverly remembers what account a file is mapped to. 

Since we already link any given input format to a specific account, we should also link each input format to a specific budget. 

Downsite is that the API feature can't yet handle if one format is used for more than one budget. This is already the case: it can't handle if a format is used for more than one budget. So it's no worse than we already are. 

Perhaps later the program should not only ask what budget&account to map to, but also whether to remember this - so the user has the choice to say no if they really have the same format for several budgets/accounts.

nocalla:
**Reference Issue:**
Fixes #275

**Description**
Budgets are now paired on a per-bank basis, allowing for multiple budgets to be handled without input each time. Unfortunately, it's not possible to support use of the same bank across multiple budgets.

- Remove concept of "default budget" unless there is only one budget in YNAB
- New budget and account selection logic
- New posting logic, as posting data requires the budget_id to be provided separately, so each budget's transactions are posted in a separate API action


nocalla:
I'm going to write another test to make sure that the transaction data is being properly created before testing it on an actual account! All looks promising so far though!

test/test_YNAB_API.py
```diff
@@ -43,24 +43,24 @@ def __init__(self, config_object, transactions=None):
...
+                # generate our list of budgets
+                budget_ids = self.list_budgets()
+                # if there's only one budget, silently set a default budget
+                if len(budget_ids) == 1:
+                    self.budget_id = budget_ids[0]
+
+                budget_t_data = self.process_transactions(transaction_data)
+                for budget in budget_ids:
+                    if budget_t_data[budget]["transactions"] != []:
+                        self.post_transactions(budget_t_data[budget])
```
nocalla: Need a new test to make sure the budget to transaction pairing is appropriate.

nocalla: I didn't do this but am going to merge and fix problems if they arise.

PR #317, opened 2020-01-09
Update core code so JLP Card UK doesn't need a separate plugin

clancychilds:
**New Pull Request**

**Description**
Previously I had a plugin that made things work for the John Lewis Partnership card in bank2ynab but ran into some problems with it in the API integration. @nocalla pointed out that some of the stuff I was doing in the plugin could probably be done with some changes to the regular functionality, thus negating the need for a separate plugin. This is the result of those changes.

1. I have added a config flag "Discard Invalid Dates" that, when set, will remove any rows with invalid dates. (This is because JLP sometimes outputs dates of transactions that are "Pending")
2. I have added a check that, if it sees an outflow has an explicit `+` sign before it, it makes it an inflow. This is similar to the existing logic that changes inflows with '-' signs to outflows. If this isn't acceptable, the other change would be to expand the outflow/inflow flag option, which is already a bit convoluted.
3. Added a check that cleans the outflow and inflow fields of anything other than digits and `.`
4. Added a bunch of tests for the above items.
5. Took the opportunity to fix some flake8 line-too-long errors.
6. Removed the old JLP Card plugin


clancychilds:
FWIW I think the only reason that automatic linter check is failing is because it is looking for this branch in this repo rather than in my forked repo.

bank2ynab.conf
```diff
@@ -12,6 +12,7 @@ Use Regex for Filename = False
 Source CSV Delimiter = ,
 Header Rows = 1
 Footer Rows = 0
+Discard Invalid Dates = False
```
nocalla: This is interesting. I can't think of any reason where we *wouldn't* want to discard invalid dates! I wonder is this flag necessary.

bank2ynab.conf
```diff
@@ -662,10 +663,10 @@ Input Columns = Date,Payee,Inflow,skip
 # MRS_JANE_SMITH_01-11-2019_01-12-2019.csv
 Source Filename Pattern = [A-Z_]*_[0-9]{2}-[0-9]{2}-[0-9]{4}_[0-9]{2}-[0-9]{2}-[0-9]{4}
 Use Regex for Filename = True
-Plugin = JLP_Card_UK
 Input Columns = Date,Payee,Outflow,skip
 Header Rows = 1
 Date Format = %d-%b-%Y
+Discard Invalid Dates = True
```
nocalla: See previous comment about whether this is a necessary flag.

bank2ynab.py
```diff
@@ -316,6 +319,7 @@ def fix_conf_params(conf_obj, section_name):
         "header_rows": ["Header Rows", False, ""],
         "footer_rows": ["Footer Rows", False, ""],
         "date_format": ["Date Format", False, ""],
+        "discard_invalid_dates": ["Discard Invalid Dates", True, ""],
```
nocalla: Again, possibly not needed.

bank2ynab.py
```diff
@@ -622,14 +646,33 @@ def _fix_decimal_point(self, row):
 
         return row
 
+    def _clean_monetary_values(self, row):
```
nocalla: Ooh, this is a potentially dangerous function. You have it in the right place in the code though - in theory, any currency using apostrophes as decimal points would have been fixed by this point.

bank2ynab.py
```diff
@@ -622,14 +646,33 @@ def _fix_decimal_point(self, row):
...
+        """ if our row doesn't have an inflow, outflow or a valid date,
+        mark as invalid
         :param row: list of values
         """
         inflow_index = self.config["output_columns"].index("Inflow")
         outflow_index = self.config["output_columns"].index("Outflow")
         if row[inflow_index] == "" and row[outflow_index] == "":
             return False
+        # check that date matches YYYY-MM-DD format
+        if self.config["discard_invalid_dates"]:
```
nocalla: I think we can remove `if self.config["discard_invalid_dates"]:` and the relevant config param- we should never allow a row with an invalid date. We should include an error message telling anybody with a failing bank the link to go to with info on date formats.

nocalla:
> FWIW I think the only reason that automatic linter check is failing is because it is looking for this branch in this repo rather than in my forked repo.

Agreed, that seems to be it. Interesting limitation to the workflow.

bank2ynab.py
clancychilds: Yeah. I made sure it was after the conversion of any `,` to `.` By this point it should be safe. (That's what unittests are for!)

bank2ynab.py
clancychilds: OK. I'm pulling it out, but I am now going to need to fix a few test cases that are failing because the dates aren't in the right format. Will try to do that in a bit.

test-data/test.conf
```diff
@@ -82,7 +82,7 @@ YNAB Account ID = Test Budget ID||Test Account ID
 
 [test_jlp_card]
 Use Regex For Filename = True
-Plugin = JLP_Card_UK
 Input Columns = Date,Payee,Outflow,skip
 Header Rows = 1
+Discard Invalid Dates = True
```
nocalla: Don't forget this one!

clancychilds:
Thanks @nocalla . I have made the suggested changes, plus I had to update some of the other tests that were now failing because of the date validation stuff in there now.

nocalla:
Great work, @clancychilds - this should mean less maintenance going forward! One fewer plugin to worry about! Thanks for the work put in here.

Issue #328, opened 2020-02-15
Support Fiducia Banks (Volksbank, Sparda-Bank, PSD, ...)


M4a1x:
**New bank format**
- Full name of the bank: Fiducia Banks (Volksbank, Sparda-Bank, PSD, ...)
- Country (2-character country code): DE

**NOTE**: The Fiducia IT is a company responsible for the IT infrastructure of a lot of german "Genossenschaftsbanken". These include the above mentioned ones and more. As I currently only have access to "Sparda-Bank Südwest" the export format might differ with differ slightly with different banks.
I also found [this ynab converter project](https://bitbucket.org/ctheune/ynab-bank-imports/src/default/src/ynab_bank_import/fiducia.py) also implementing fiducia, suggesting that indeed it is the same format for all banks.

I'm planning on issuing a pull request on this.

```
[DE Fiducia (Volksbank, Sparda-Bank, BBBank, PSD Bank, ...)]
# Only tested with Sparda-Bank
Use Regex For Filename = True
Source Filename Pattern = Umsaetze_[A-Z]{2}\d{20}_\d{4}.\d{2}.\d{2}
Source CSV Delimiter = ;
Header Rows = 12
Footer Rows = 3
Input Columns = Date,skip,skip,Payee,skip,skip,skip,skip,Memo,skip,skip,Inflow,CDFlag
Date Format = %d.%m.%Y
Inflow or Outflow Indicator = 12,H,S
```

M4a1x:
How would I go about adding tests for it with the test file I posted in #330 as you suggested? 
I found test_read_data in the `test_B2YBank.py`. Am I correct in that I would have to add the .csv to `test-data` and copy the configuration for the bank from `bank2ynab.conf` to `test-data/test.conf`?

nocalla:
All you need to do is add a sample anonymised CSV file to the test-data folder. We don't currently have automated testing of all the bank formats, but it could be a helpful reference for someone in the future.

PR #329, opened 2020-02-15
Fix csvreader line numbering.

M4a1x:
**New Pull Request**
Noticed (after some debugging...) while implementing #328 that only 2 lines would get converter and found an error in the line numbering.

**Description**
line_num is not equals the rows of the csv. It basically counts \n characters. This fails, when entries contain newlines.
See https://docs.python.org/3.7/library/csv.html#csv.csvreader.line_num

**Changes**
It's a simple and small change, since the csvreader is already used in a for loop. Instead of asking the underlying stream for a line number, just count the loops (the same approach as the lines above for getting the total number of rows).
This fixes especially any issues with line numbering an headers.

### DIFF
--- a/bank2ynab.py
+++ b/bank2ynab.py
@@ -445,7 +445,7 @@ class B2YBank(object):
             # make each row of our new transaction file
             for line, row in enumerate(transaction_reader):
                 # skip header & footer rows
-                if header_rows < line <= (row_count - footer_rows):
+                if header_rows <= line <= (row_count - footer_rows):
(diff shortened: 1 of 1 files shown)


M4a1x:
I noticed that I have to adjust the if clause as well. Should I squash & force push the commits, or is it ok like this?

The linter is still failing, but I don't know how to make it pull from my fork instead of bank2ynab. I guess something has to be adjusted in the workflow? Sorry, I'm not familiar with travis.

PR #362, opened 2020-09-01
Adding possibility of lower/upper case extension

ProfileID:
I wanted to import an export from Commerzbank which got downloaded as ".csv".
The default file extension is ".CSV", which caused the problem that the file was not found or was ignored.

How about searching for the extension in lower case as well as upper case?
In my commit the lower and upper function is used to better recognize the file extension.

### DIFF
--- a/bank2ynab/bank_process.py
+++ b/bank2ynab/bank_process.py
@@ -77,7 +77,7 @@ class B2YBank(object):
                 files = [
                     join(path, f)
                     for f in directory_list
-                    if f.endswith(ext)
+                    if f.endswith(ext) or f.endswith(ext.lower()) or f.endswith(ext.upper())
                     if f.startswith(file_pattern)
                     if prefix not in f
                 ]


nocalla:
Good catch on this problem. I have a different solution though that I think is a bit more straightforward.

`if f.lower().endswith(ext)`

bank2ynab/bank_process.py
```diff
@@ -77,7 +77,7 @@ def get_files(self):
                 files = [
                     join(path, f)
                     for f in directory_list
-                    if f.endswith(ext)
+                    if f.endswith(ext) or f.endswith(ext.lower()) or f.endswith(ext.upper())
```
nocalla, line 80: `if f.lower().endswith(ext) `
as a replacement for the whole of line 80 here should work in theory. It converts the entire filename to lower case and then runs the comparison against that.

nocalla:
Fixed this a different way by amending the bank format to have a lower case extension.

PR #367, opened 2020-10-26
Adds support for new NL ING format

### LINKED ISSUE #365: Error NL ING Checking 2020: IndexError when when running comparison on indicator_col for CD_flag
Dear community,

After having worked perfectly for a while. The last couple of weeks it is not working anymore. I just downloaded the latest developer code, but that does also not solve the issue.

This is the error I am getting:

```
XXX-MBP:bank2ynab-develop XXusernameXX$ python3 ./bank2ynab
INFO: 
Parsing input file:  /Users/username/Downloads/NL00INGB0000000000_06-08-2020_26-08-2020.csv (format: NL ING Checking 2020)
INFO: Using encoding ascii with confidence 1.0
INFO: Using encoding ascii with confidence 1.0
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/runpy.py", line 192, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "./bank2ynab/__main__.py", line 19, in <module>
    b2y.run()
  File "./bank2ynab/bank_process.py", line 382, in run
    output = bank.read_data(src_file)
  File "./bank2ynab/bank_process.py", line 126, in read_data
    row = self._cd_flag_process(row, cd_flags)
  File "./bank2ynab/bank_process.py", line 305, in _cd_flag_process
    if row[indicator_col] == outflow_flag:
IndexError: list index out of range
XXX-MBP:bank2ynab-develop XXusernameXX$ 
```

*for privacy reasons I censored my username and banking number (changed to zero's and X's).*

Please advice.

SliceOLife:
**Reference Issue:**
Fixes #365 

**Description**
Apparently the Dutch bank ING added a new format that changes the delimiter to a semicolon and adds two new (useless) fields.

This adds a new variant of the NL ING checking 2020 entry to support the semicolon-delimited version. The old version is also still available to download on the ING website so I left that in as well.

### DIFF
--- a/bank2ynab.conf
+++ b/bank2ynab.conf
@@ -498,18 +498,6 @@ Use Regex for Filename = True
 Source Filename Pattern = NL[0-9]{2}INGB[0-9]*_[0-9]{2}-[0-9]{2}-[0-9]{4}_[0-9]{2}-[0-9]{2}-[0-9]{4}
 Header Rows = 1
 Footer Rows = 0
-# Datum,"Naam / Omschrijving","Rekening","Tegenrekening","Code","Af Bij","Bedrag (EUR)","MutatieSoort","Mededelingen"
-Input Columns = Date,Payee,skip,skip,skip,CDFlag,Inflow,skip,Memo
-Date Format = %Y%m%d
-Inflow or Outflow Indicator = 5,Bij,Af
-
-[NL ING Checking 2020 2]
-# source: survey response #53, 20200116 
-# NL04INGB0704279991_01-01-2019_11-01-2020.csv
-Use Regex for Filename = True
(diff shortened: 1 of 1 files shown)


nocalla:
Great, thanks for putting in the time on this. Would you do me a favour and simply overwrite the other version of the bank format instead of adding another version of it - I assume the filename pattern is the same, so having two formats keyed to the same pattern will cause problems.

nocalla (approved):
Great, thanks.

Issue #383, opened 2021-04-09
Handle multiple accounts from different banks


jtpedersen:
I made fork that is able to handle multiple accounts from different banks.

It works by: 
1. Generate a matching tree structure to the budgets on YNAB
2. The user (me) puts csv files in the folders matching the accounts
3. The script walks throug each folder in the tree and determines the bank type (json file in folder) and posts the data to the ynab api

The state is that it works for me and I think that is a nifty addition usage of bank2ynab.  I did some refactorings on various api parts to understand the code and to force it to work for this use case.

If there is any interest in getting this into mainline I can try to clean up the commits etc?

https://github.com/jtpedersen/bank2ynab/tree/multi_accounts

I wrote a bit on gitter but got no response and now I just added the different banks feature so...

jtpedersen:
This changes the general flow of bank2ynab so the mapping between a csv file and the corresponding bank and account no longer is fixed and determined by a regex on the filename. I had a bank where i could change the filename and had mulitiple accounts and could not understand/make the default work. This is relationsship is established by the relative path to the file (for account) and meta-information for the bank identifier.

To illustrate

```
:~/src/bank2ynab$ tree imports/
imports/
└── My Budget
    ├── AL-budget
    │   └── account_info.json
    ├── AL-Helle
    │   └── account_info.json
    ├── Al-jacob
    │   └── account_info.json
    ├── AL-Jacob-Master
    │   └── account_info.json
    ├── AL-MadKontoen
    │   └── account_info.json
    ├── MidSpar-bolig
    │   └── account_info.json
    └── MidSpar-forbrug
        └── account_info.json

8 directories, 7 files
```
Where the content of account_info.json is a simple json with contents like `{"bank_id": "DK Portalbank"}` but that information is now specific for each account.

The initial structure is build by querying the ynab api for the budgets and accounts.

Any csv file put into a subfolder til when running the bank2ynab program will then be exported the YNAB and then have the file moved to a history folder, so it will not be parsed again and there is a record of what has been imported. Having this may help find errors and/or could in future be extended to ensure that each transaction is only uploaded once.

nocalla:
As mentioned on #386, I don't think this fits with the workflow goal of the project. Marking as won't fix. Thanks again.

PR #386, opened 2021-04-09
[WIP] Multi accounts

jtpedersen:
Issue #383

Support multi accounts from different banks


bank2ynab.conf
```diff
@@ -2,7 +2,7 @@
 
 [DEFAULT]
 # Specify where you save your downloaded CSV files:
-Source Path = %USERPROFILE%\Downloads
+Source Path = /home/jacob/Downloads/
```
nocalla, line 5: I think we want to keep the generic user here. Not everyone has the same first name.

bank2ynab.conf
```diff
@@ -13,20 +13,21 @@ Source CSV Delimiter = ,
...
+# (see https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes for date format strings)
 Date Format =
 Inflow or Outflow Indicator =
 # Output file formatting
 Output Columns = Date,Payee,Category,Memo,Outflow,Inflow
 Output Filename Prefix = fixed_
 Use Payee for Memo = True
 # Post-processing
-Delete Source File = True
+Delete Source File = False
```
nocalla, line 24: Why are we changing this default?

bank2ynab.conf
```diff
@@ -13,20 +13,21 @@ Source CSV Delimiter = ,
...
 # Plugins
 Plugin =
 # YNAB API Stuff - don't modify values here, modify in user_configuration.conf
-YNAB API Access Token =
-YNAB Account ID =
+#ffc24c41e613dec1e9eb72bfac15837c7b3014635963f952a01a44d86dac393c
```
nocalla, line 28: Is this your personal API key??

nocalla:
I'm confused as to what this actually does. The first time we do a write to the API, there's a dialogue asking which account you want to match each bank to, which means you can just dump all your files in one directory and run the script. This seems easier to me - am I missing something?

bank2ynab.conf
```diff
@@ -13,20 +13,21 @@ Source CSV Delimiter = ,
 Header Rows = 1
 Footer Rows = 0
 Input Columns = Date,Payee,Outflow,Inflow,Running Balance
-# (see https://docs.python.org/2/library/datetime.html#id1 for date format strings)
+# (see https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes for date format strings)
```
nocalla, line 16: This is weird - `develop` already uses this Python 3 documentation link, why is it showing as a change?

bank2ynab.conf
```diff
@@ -340,12 +351,23 @@ Input Columns = skip,Date,skip,skip,Memo,skip,skip,skip,skip,skip,skip,Payee,ski
...
 # Tested Banks: Arbejdernes Landsbank, Merkur Andelskasse, DK Handelsbanken,
 # add more here..
-Source Filename Pattern = Posteringer
+Source Filename Pattern = Al
 Source CSV Delimiter = ;
 Header Rows = 0
 Input Columns = Date,Payee,Inflow,Running Balance
 Date Format = %d-%m-%Y
 
+[DK Portalbank]
```
nocalla, line 360: This is generating a conflict  - can you remove and pull from `develop` please?

bank2ynab/__main__.py
```diff
@@ -1,22 +1,85 @@
...
+    """
+    root = get_import_root()
+    for budget in budgets:
+        for account in budget.accounts:
+            folder = os.path.join(root, budget.name, account.name)
+            os.makedirs(folder, exist_ok=True)
+            config = os.path.join(folder, CONFIG_FILE)
+            if not os.path.exists(config):
+                with open(config, "w") as f:
+                    json.dump({"bank_id": "DK Bankernes EDB Central"}, f)
```
nocalla, line 34: Should a specific bank be named here?

bank2ynab/__main__.py
```diff
@@ -1,22 +1,85 @@
...
+
+
+Job = namedtuple("Job", ("budget", "account", "filename", "bank_id"))
+
+
+def move_to_history_folder(job):
```
nocalla, line 40: I don't see the need for a history folder - is that not the function that YNAB is fulfilling? In any case, I feel like it's a separate feature.

bank2ynab/__main__.py
```diff
@@ -1,22 +1,85 @@
...
+    os.rename(job.filename, os.path.join(folder, name))
+
+
+def process_import_tree(b2y, budgets, api):
+    root = get_import_root()
+    jobs = []
+    for budget in budgets:
+        for account in budget.accounts:
+            account_path = os.path.join(root, budget.name, account.name)
+            bank_id = "DK Bankernes EDB Central"
```
nocalla, line 55: A specific bank name again.

jtpedersen:
Yes there is a lot of cruft that should be removed and even things that will be fixed by rebase on a recent version of develop. This is why it was marked as WIP "Work In Progress"

I made this to suite my personal workflow and it "works for me" but it does change the general flow of bank2ynab and as such it may not be that you want any of the changes making it into  bank2ynab. Which is why i have not done the necessary cleanup, iff there is any interest in this feature or parts of it, i will clean it up and make a nice PR. But if it is not fitting into the current plan or flow then I can just keep my "weird setup" in my own fork.

bank2ynab/__main__.py
jtpedersen: I found it to be a nice feature that my csvs does not disapear, upon importing and that i know when they were imported. But yes the data should be in ynab when the script is done.

bank2ynab.conf
jtpedersen: Ye the whole PR would need a rebase on a recent origin/develop to work. :+1:

bank2ynab.conf
jtpedersen: it was accidentially commited and revoked some time ago.

jtpedersen: https://github.com/bank2ynab/bank2ynab/pull/386#commitcomment-47634433

bank2ynab.conf
jtpedersen: :+1: then again not everybody uses windows so it should perhaps be some standard pythonpath https://stackoverflow.com/questions/35851281/python-finding-the-users-downloads-folder seems to suggest that it is possible to do.

jtpedersen:
I wrote a more extensive comment on the idea / features on the associated issue #383

bank2ynab.conf
nocalla: That's what the find_directory function in b2y_utilities does if needed. If Source Path is blank, it uses that.

nocalla:
I like some of the refactoring you've done and it's good to see what doesn't make sense to a fresh pair of eyes, but I think this feature doesn't quite fit with the "one-click" goal of the project since you have to sort into folders beforehand. Thanks for the contribution though.

jtpedersen:
Super. Thank you for taking time to look at it. Please do cherry pick anything you do find useful. Have a nice day.

PR #403, opened 2021-11-16
Add YNAB imported vs skipped metrics

ThomDietrich:
A quick fix to make the script a bit more chatty. Imho definitely an area this great piece of code can have a bit more of ;)


nocalla:
Good idea!

Issue #395, opened 2021-06-14
Uploading transactions to YNAB silently fails


FlipEverything:
**Describe the bug**
File processing is done, connecting to YNAB API is done, but nothing is actually imported to YNAB. No error message is shown. If I manually import the `fixed_export.csv` via YNAB web UI than it works. It was fine for 2 weeks but now it does not work.

**What did you EXPECT to happen?**
When I open YNAB there are the imported transactions.

**What ACTUALLY happened?**
The program said the upload was successful but the transactions are missing from YNAB.

**What did you DO? (steps to reproduce)**
1. create `export.csv`
2. run `python3 ./bank2ynab`

**What's your software environment?**
- Script language: Python 3.9.4
- Operating system: macOS
- OS version: 11.2

**Can you provide other helpful information?**
```python
INFO:
Parsing input file:  szamlatortenet_20000101_29991231.csv (format: HU K&H)
INFO: Using encoding utf-8 with confidence 0.99
INFO: Using encoding utf-8 with confidence 0.99
INFO: Parsed 2 lines
INFO: Writing output file: fixed_szamlatortenet_20000101_29991231.csv
INFO:
Done! 1 files processed.

INFO: Connecting to YNAB API...
INFO: Processing transactions...
INFO: Previously-saved account for HU K&H found.
INFO: Uploading transactions to YNAB...
INFO: Success!
```

FlipEverything:
Any updates on this? Could you please help me to debug this?

nocalla:
Hmm. This looks like a problem with account matching. Can you try opening the `user_configuration.conf` file and deleting the account ID associated with the bank? Then retry and post the output.

FlipEverything:
It's the same. I double-checked the ID and it's correct. Sometimes it works with the exact same ID.

`ynab account id = aXXXXd87-7X4e-44Xf-89fd-dfXXXXX52a45||7XXXe409-6X99-4X10-a1f5-6e6XXXXX4e6a`

I contacted YNAB if they can help debug this from their side, but they won't:

> Thanks for letting us know about this!
> 
> Since it's not officially supported by us and importing the file directly in YNAB is working fine, you'll need to reach out to the developers over at bank2ynab about that.
> 
> Let us know if we can help with anything else!

nocalla:
[...]

PR #406, opened 2021-11-21
Migration to Pandas for data handling

### LINKED ISSUE #273: More helpful API error messages
The API error messages are rather generic and I think we can improve them - in particular the error 401 `API access token missing, invalid, revoked, or expired` ([link to code line](https://github.com/bank2ynab/bank2ynab/blob/develop/bank2ynab.py#L883)) because users are very likely to encounter this error on their first try of the new API-enabled bank2ynab.

The error messages should be longer. In particular, they should not only describe the problem but also describe possible solutions and/or link to our wiki and/or the YNAB (API?) help pages for details. 

A first step is to provide a little guidance right in the error message.

**Additional context**
Relevant link for the 401 error: https://api.youneedabudget.com/#personal-access-tokens 

I prefer to link directly to YNAB pages because then we don't have to maintain our own wiki page and run the risk of it being outdated. OTOH a wiki page can contain anything we want!

### LINKED ISSUE #401: Code refactor & migrate parsing functionality to Pandas
Our code is a bit spaghettified due to the gradual expansion of features. I propose migrating the code to use the [`Pandas`](https://pandas.pydata.org/) 3rd party library as well as conducting a bit of a refactor. Considering we are already using the `requests` 3rd party library, an additional dependency shouldn't be too disruptive.

- Migrate our input parsing to the Pandas library. This has the advantage of being able to handle many types of input, so could potentially boost our project's functionality
-  Review the API segment of the code, implementing more robust handling of API error codes (#273)

I'm not considering leaving the code as-is, as it's giving me headaches to try to read or fix anything!

nocalla:
**Reference Issue:**
Fixes #401

**Description**
Major overhaul of the core workflow\! I definitely got carried away with this one.
- Remove lingering Python 2 support workarounds
- Move to Pandas library for processing data
- Add a Currency Conversion Factor config flag to allow 
- Remove RaiffeisenRCM plugin in favour of the Currency Conversion Factor config flag
- Refactor division of code across modules 
- Improve YNAB error responses (fixes #273)
- Improve 403.4 API error response thanks to #395!
- Change github linting workflow to just call Black (hopefully will fix the issues with Pull Requests from forks)
- Added typehints throughout (started using VScode, so the typing helped a lot with this refactor)
- Added a generic PDF plugin
- Added a Plugin Arguments config flag to allow for more generic plugins that can be customised with arguments instead of being fully custom-written for each bank


bank2ynab/bank_process.py
```diff
@@ -95,58 +104,67 @@ def read_data(self, file_path):
...
+            skipinitialspace=True,  # skip space after delimiter
+            header=None,  # don't set column headers initially
+            skiprows=header_rows,  # skip header rows
+            skipfooter=footer_rows,  # skip footer rows
+            skip_blank_lines=True,  # skip blank lines
+            encoding=encod,
+            memory_map=True,  # access file object directly - no I/O overhead
+            engine="python",
+        )
+
```
nocalla, line 133: I think a split is needed here to separate reading and parsing the data.

nocalla: Maybe even a separate class for TransactionData and BankData? Not sure where to draw the line for the structure. This class has too many responsibilities right now.

bank2ynab/__main__.py
```diff
@@ -17,6 +17,6 @@
     else:
         b2y = Bank2Ynab(config)
         b2y.run()
-        api = YNAB_API(config)
-        if b2y.transaction_data:
-            api.run(b2y.transaction_data)
+        # api = YNAB_API(config) # DEBUG: disabled
+        """ if b2y.transaction_data: # DEBUG - disabled while testing
+            api.run(b2y.transaction_data) """
```
nocalla, line 22: 1. try get configuration files (main & user-config [maybe payees mapping too])
    - Gracefully fail if there's no config file
2. for each section in the configuration file, create a bank class with the relevant parameters saved
3. create a list of bank classes
4. for each bank class in the list
    - find all matching files
    - read data into a dataframe from each file
    - attempt to parse & clean the data
    - validate data
    - write data to output csv file
5. output all transaction data to YNAB api

bank2ynab/bank_iterator.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def build_bank(self, bank_config: dict):
+        # TODO mostly commented out for now as plugins need to be fixed
+        """Factory method loading the correct class
+        for a given configuration."""
+        # plugin_module = bank_config.get("plugin", None)
+        """ if plugin_module:
+            p_mod = importlib.import_module("plugins.{}".format(plugin_module))
+            if not hasattr(p_mod, "build_bank"):
+                s = (
```
nocalla, line 32: API connection needed here

bank2ynab/dataframe_handler.py
```diff
@@ -0,0 +1,309 @@
...
+        (https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior)
+        :type date_format: str
+        :return: modified dataframe
+        :rtype: Series
+        """
+        date_series = pd.to_datetime(
+            date_series,
+            format=date_format,
+            infer_datetime_format=True,
+            errors="coerce",
```
nocalla, line 295: TODO - add the columns required by the YNAB API

bank2ynab/dataframe_handler.py
```diff
@@ -0,0 +1,309 @@
...
+        # fill in null values with 0
+        num_series.fillna(value=0, inplace=True)
+        # currency conversion if multiplier specified
+        num_series = num_series.astype(float) / self.currency_fix
+        return num_series.astype(float)
+
+    def _remove_invalid_rows(self) -> None:
+        """
+        Removes invalid rows from dataframe.
+        An invalid row is one which does not have a date
```
nocalla, line 247: Rework this to use the dataframe.

bank2ynab/dataframe_handler.py
```diff
@@ -0,0 +1,309 @@
...
+        :param fill_memo: boolean to check
+        :type fill_memo: bool
+        :return: None
+        """
+        if fill_memo:
+            self.df["Memo"].fillna(self.df["Payee"], inplace=True)
+
+    def _auto_payee(self) -> None:
+        """
+        if Payee is blank, fill with contents of Memo column
```
nocalla, line 271: Rework this to use the dataframe.

bank2ynab/dataframe_handler.py
nocalla, line 295: This function may not actually be needed any more.

bank2ynab/dataframe_handler.py
```diff
@@ -0,0 +1,309 @@
+import logging
+
+import pandas as pd
+
+
+class DataframeHandler:
+    # TODO - integrate payee mapping in this class
+
```
nocalla, line 8: This is where @ThomDietrich 's payee transformation should integrate.

bank2ynab/ynab_api.py
```diff
@@ -1,29 +1,38 @@
...
     """
     uses Personal Access Token stored in user_configuration.conf
     (note for devs: be careful not to accidentally share API access token!)
     """
 
-    def __init__(self, config_object, transactions=None):
+    def __init__(
+        self, config_object: ConfigHandler, transactions=None
+    ) -> None:
         self.transactions = []
```
nocalla, line 27: Can passing the ConfigHandler be avoided? Can we pass the parameters explicitly instead?

bank2ynab/ynab_api.py
```diff
@@ -1,29 +1,38 @@
...
-import b2y_utilities
+from config_handler import ConfigHandler
 
 # configure our logger
 logging.basicConfig(format="%(levelname)s: %(message)s", level=logging.INFO)
 
 
-class YNAB_API(object):  # in progress (2)
+class YNAB_API:
+    # TODO - revise docstring
```
nocalla, line 14: This class is too unwieldy and needs to be broken up - maybe handle the API responses elsewhere?

PR #412, opened 2022-01-08
Parse Attributes in Memo Field

### LINKED ISSUE #411: Extraction of transaction information from the memo
**Other feature request**

Some of my bank accounts put all of the useful information (actual transaction date, description, payee, etc) in the memo field. This means that using bank2ynab as it stands results in data in YNAB like:

> Date: 2021-12-22 (the settle date not transaction date)
> Payee: ""
> Memo: PAYMENT VIA MAESTRO                  22-12 21-12-2021 AT 17.00 TIME MCDONALDS AEROP LISB 1500-284 WITH APPLE PAY XX XXXX X6 VIRTUAL CARD FOR CONTACTLESS: XXXX X888 9 CARDHOLDER: NAME HERE

I'd like a built in way (perhaps via a .conf file) to extract information from this memo field, it could be perhaps a regex that documents the format, then information from the memo, rather than the CSV's other rows, are used in YNAB.

Manually, or automatically before bank2ynab processing, clean up all the data.

**Additional context**

The biggest difficulty is not all banks seem to do this, and they'll all have widely different formats. Whatever solution is implemented it needs to avoid being brittle.

Would you accept a pull request implementing such a feature? (if I have the time).

Pezmc:
**Reference Issue:**
Fixes #411 

**Description**
Adds a plugin that can parse the memo field to extract useful information that would otherwise be lost.

The plugin and options must be added to a banks config, like so:

```conf
Source Filename Pattern = PatternHere
Input Columns = Date,Inflow,skip,Memo,skip,Payee
Plugin = parse_from_memo
Plugin Arguments =
	(?P<date>\d{2}-\d{2}-\d{4}) AT (?P<time>\d{2}\.\d{2}) WITH  (?P<memo>.+)
	(?P<date>\d{2}-\d{2}-\d{4}) TO  (?P<payee>.+)
```

The `Plugin Arguments` field is a list of regular expressions with named groups.
Valid group names are `payee`, `memo`, `date`, `time`, `purchaser`

I also fixed some bugs in the develop branch (as of last night) that needed resolving in order to test the plugin, see the commit messages for full explanation.

### DIFF
--- a/bank2ynab/bank_handler.py
+++ b/bank2ynab/bank_handler.py
@@ -72,6 +72,7 @@ class BankHandler:
                 logging.info(
                     f"No output data from this file for this bank. ({e})"
                 )
+                logging.debug(traceback.format_exc())
             else:
                 # make sure our data is not blank before writing
                 if not df_handler.df.empty:
--- a/bank2ynab/bank_handler.py
+++ b/bank2ynab/bank_handler.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import traceback
 from os.path import basename, dirname, isfile, join
 from typing import Any
 
--- a/bank2ynab/bank_handler.py
+++ b/bank2ynab/bank_handler.py
@@ -1,11 +1,14 @@
 import logging
+
+from os import path
 import os
 import traceback
 from os.path import basename, dirname, isfile, join
 from typing import Any
 
-from dataframe_handler import DataframeHandler, combine_dfs
-from transactionfile_reader import detect_encoding, get_files
+import dataframe_handler
+from dataframe_handler import DataframeHandler
+import transactionfile_reader
 
 
 class BankHandler:
@@ -28,7 +31,7 @@ class BankHandler:
         self.transaction_list: list[dict] = list()
 
     def run(self) -> None:
-        matching_files = get_files(
+        matching_files = transactionfile_reader.get_files(
             name=self.config_dict["bank_name"],
             file_pattern=self.config_dict["input_filename"],
             try_path=self.config_dict["path"],
@@ -48,7 +51,7 @@ class BankHandler:
                     plugin_args=self.config_dict["plugin_args"],
                 )
                 # get file's encoding
-                src_encod = detect_encoding(src_file)
+                src_encod = transactionfile_reader.detect_encoding(src_file)
                 # create our base dataframe
 
                 df_handler = DataframeHandler()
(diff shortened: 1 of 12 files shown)


bank2ynab/bank_handler.py
```diff
@@ -72,6 +73,7 @@ def run(self) -> None:
                 logging.info(
                     f"No output data from this file for this bank. ({e})"
                 )
+                logging.debug(traceback.format_exc())
```
Pezmc, line 76: Without this, errors in plugins like the `I/O operation on closed file` bug are just silently swallowed

bank2ynab/config_handler.py
```diff
@@ -101,7 +101,7 @@ def fix_conf_params(self, section: str) -> dict[str, Any]:
             ),
             "plugin": self.get_config_line_str(section, "Plugin"),
             "plugin_args": self.get_config_line_lst(
-                section, "Plugin Arguments", ","
+                section, "Plugin Arguments", "\n"
```
Pezmc, line 104: No plugin uses the `Plugin Arguments` option yet, so this seemed safe to update.

bank2ynab/dataframe_handler.py
```diff
@@ -123,18 +123,20 @@ def read_csv(
...
-        names=[],  # don't set column headers initially
-        skiprows=header_rows,  # skip header rows
-        skipfooter=footer_rows,  # skip footer rows
-        skip_blank_lines=True,  # skip blank lines
-        encoding=encod,
-        memory_map=True,  # access file object directly - no I/O overhead
-        engine="python",
-    )
+
+    with open(file_path, 'rU') as csv_file:
```
Pezmc, line 127: I was experiencing an issue with pandas.read_csv throwing `new-line character seen in unquoted field - do you need to open the file in universal-newline mode?` seemingly because my bank was formatting files using unix style line endings, instead of windows. Adding the `rU` universal line endings flag and using a file handle resolves this.
`

bank2ynab/plugins/parse_from_memo.py
```diff
@@ -0,0 +1,138 @@
...
+
+        # Generate output path
+        new_path = get_output_path(
+            input_path=file_path,
+            prefix=f"memo_parsed_csv_{self.config_dict['bank_name']}_",
+            ext=".csv",
+        )
+
+        # Write the dataframe to output file
+        new_df.to_csv(new_path,
```
Pezmc, line 70: Having to read from the CSV, to write into a CSV and re-load is a less than ideal plugin interface. I originally built against master and manipulated the data _after_ the BankHandler had read in the CSV; however it looks like this is the best way with the "new" plugin API.

nocalla: I know, it's a bit clunky, but the logic was that a plugin should exist to get the file into a "base" state where all normal operations can then be applied on it. It just seemed like the most universal approach.

nocalla:
That's great! Thanks for the contribution (and the bug fixing!).

PR #424, opened 2022-07-25
fix(bank2ynab.conf): skip new columns for Danske

sadikaya:
**New Pull Request**

**Reference Issue:**
No existing issue.

**Description**
In the output data for Danske Bank accounts, there is now two new columns, "Kategori" and "Underkatagori".
I've updated the configuration to skip both of these values, so that it can parse the csv and map to the correct columns.


nocalla:
Great, thanks for the contribution!

PR #427, opened 2022-12-04
Add support for two Swiss banks

divadvo:
**New Pull Request**

**Reference Issue:**
*Type # followed by the issue number this pull request pertains to*

**Description**

I've added support for SwissCard bank.
And I've added support for new csv format for checking accounts from UBS.


nocalla:
Great, thanks for the contribution.

PR #429, opened 2022-12-26
Updated .csv for SE Länsförsäkringar

alanlahoni:
**New Pull Request**

**Reference Issue:**
No existing issue.

**Description**
The output .csv for SE Länsförsäkringar has changed. The columns for "Kontonummer";"Kontonamn";"";"Saldo";"Tillgängligt belopp" is now in the header, which also results in 4 Header Rows instead of 1. The Date Format has also changed and is now %Y-%m-%d instead %d/%m/%y.

Also "Transaktionstyp" should be used as Memo instead of skipped.

I have updated the bank2ynab.conf to reflect these changes.


nocalla:
Thanks for the contribution!

PR #438, opened 2023-03-13
OCBC: strip quotes, update header rows

ExalDraen:
**Description**

Updates and fixes to handle the latest transaction history exports from OCBC bank in Singapore. 

- Strip single quotes (`'`) from lines as they are being processed by the OCBC plugin
- Update the OCBC import configurations to:
  - update the number of header rows for credit cards (there's 7 now)
  - also import payees for accounts\
- add a few comments / explanations

### DIFF
--- a/bank2ynab/plugins/OCBC_Bank_SG.py
+++ b/bank2ynab/plugins/OCBC_Bank_SG.py
@@ -1,10 +1,11 @@
 # Plugin for handling OCBC Bank [SG] files
 
-from bank_handler import BankHandler
+from bank2ynab.bank_handler import BankHandler
 
 
 class OCBC_Bank_SG(BankHandler):
-    """Example subclass used for testing the plugin system."""
+    """Plugin for handling Oversea-Chinese Banking Corporation Singapore
+    (OCBC SG) bank files"""
 
     def __init__(self, config_dict: dict):
         """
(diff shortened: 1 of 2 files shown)


bank2ynab/plugins/OCBC_Bank_SG.py
```diff
@@ -31,6 +32,8 @@ def _preprocess_file(self, file_path, plugin_args) -> str:
         with open(file_path) as input_file:
             output_rows = []
             for rownum, row in enumerate(input_file):
+                # strip any single quotes, e.g. if payee is MCDONALD'S
```
nocalla, line 35: This is an interesting change - was this causing issues?

ExalDraen: It was, but `McDonald's` actually doesn't cause a problem - I just tried and it becomes `Mcdonald S` without the fix and `Mcdonalds` with the fix. So still an improvement but doesn't really change much.

The problem I had a while ago actually meaningfully broke the conversion - but it was a while ago and I don't have an example txn history that showcases the problem.

I can update the comment if you like?

nocalla: The clean_strings function in dataframe_handler.py replaces all non-alphanumeric characters with spaces, so it shouldn't break anything going forward. I can merge as is or you can remove this part if you want if you don't think it's necessary.

ExalDraen: Well, my preference would be to leave it as-is given that I do remember it causing issues - even if I can repro them right now. 

I noticed the README update action [failed](https://github.com/bank2ynab/bank2ynab/actions/runs/4402052212/jobs/7714965014) but it failed at the git checkout stage and I have a hard time believing my change caused that!

nocalla: No, that's definitely not you! GitHub actions just can't checkout forks.

Issue #439, opened 2023-05-20
PL Alior Bank format


psmarcin:
**New bank format**

- Full name of the bank: Alior Bank
- Country (2-character country code): PL
- Any special requirements: -

```
Source Filename Pattern = Historia_Operacji_[0-9]{4}-[0-9]{2}-[0-9]{2}_[0-9]{2}-[0-9]{2}-[0-9]{2}.csv
Source Filename Extension = .CSV
Use Regex for Filename = True
Source CSV Delimiter = ;
Header Rows = 2
Input Columns = Date,skip,Payee,skip,Memo,Inflow,skip,skip,skip,skip,skip,skip
Date Format = %d-%m-%Y
```

PR #440, opened 2023-05-20
Add PL Alior Bank format

psmarcin:
**New Pull Request**

**Reference Issue:** #439 

**Description**
New format for PL Alior Bank. 
Ran it locally for few export from different types of account (personal, savings and business types)


nocalla:
Great, thanks for the contribution.

PR #441, opened 2023-06-08
Add PL PKO BP bank

psmarcin:
**New Pull Request**

**Description**
New format for PL PKO BP. 
Ran it locally for few export from different types of account (personal, savings and business types).


nocalla:
Thanks for the contribution!

PR #442, opened 2023-07-01
Change Source CSV Delimiter for "CH ZKB Erweiterte Suche" to comma ","

fwullschleger:
**New Pull Request**

**Reference Issue:**
*Type # followed by the issue number this pull request pertains to*
- none

**Description**

**Explain the changes that this pull request makes**
*Note that any submitted code must conform to the standards set by the Black linter. Please review the automated Linting test results on the pull request after saving and make any required changes or the submitted code will not be accepted.*

- ZKB changed the CSV file one can export. The new CSV delimiter is now a comma instead of a semi-colon.

PR #443, opened 2023-07-19
Change Source CSV Delimiter back to semi-colon and add an additional row to skip to 'CH ZKB Erweiterte Suche'

fwullschleger:
**New Pull Request**

**Reference Issue:**
none

**Description**
In my previous pull request #442, I changed the Source CSV Delimiter for 'CH ZKB Erweiterte Suche' to a comma. As it seems this was a temporary change from the bank.

I had to change the Source CSV Delimiter for 'CH ZKB Erweiterte Suche' back to a semi-colon. Additionally, the bank introduced a new column to the file which can be skipped.


nocalla:
Thanks for the update.

PR #444, opened 2023-09-25
add DE Deutsche Kreditbank checking new

### LINKED ISSUE #432: No output file for this bank
**Describe the bug**
Problem with running the code. It detects the right bank and format, but does not output any file. This is in the terminal

```
(DE Deutsche Kreditbank checking)
INFO:   Opening file using encoding ISO-8859-1 (confidence 0.73)
INFO: No output data from this file for this bank. (Expected 9 fields in line 9, saw 12)
ERROR:
```

**What did you EXPECT to happen?**
An output file, no error

**What ACTUALLY happened?**
See "bug"

**What did you DO? (steps to reproduce)**
1. Download File from DKB, did not change it.
2. Run the code in terminal python bank2ynab in the folder
3. Get the error
4. Write an Issue ;)

**What's your software environment?**
 - Script language: Python 3.10.9
 - Operating system: Windows
 - OS version: 10

**Can you provide other helpful information?**
I looked into the bank2ynab.conf and saw that the criteria to look for the right file are correct

```
[DE Deutsche Kreditbank checking]
*# source: Issue #132
*# 1234567890.csv
Use Regex For Filename = True
Source Filename Pattern = [0-9]{10}
Header Rows = 7
Source CSV Delimiter = ;
Input Columns = skip,Date,Memo,Payee,skip,skip,skip,Inflow,skip,skip,skip
Date Format = %d.%m.%Y
```

I wonder why it says it saw 12 fields and expected 9. It has 11 fields, Just like in "input columns" defined. So i do not know what happens here. Sry I am some kind of a newb in programming.

martenbiehl:
**New Pull Request**

**Reference Issue:**
Fixes #432 

**Description**
DKB has a new online banking that has a different csv export format. It probably makes sense to keep the old format in for a while. They don't conflict as the file naming is different.

Only configuration changes

### DIFF
--- a/bank2ynab.conf
+++ b/bank2ynab.conf
@@ -368,6 +368,14 @@ Source CSV Delimiter = ;
 Input Columns = skip,Date,Memo,Payee,skip,skip,skip,Inflow,skip,skip,skip
 Date Format = %d.%m.%Y
 
+[DE Deutsche Kreditbank checking new]
+Use Regex For Filename = True
+Source Filename Pattern = [0-9]{2}-[0-9]{2}-[0-9]{4}_Umsatzliste_Girokonto_DE[0-9]{20}
+Header Rows = 5
+Source CSV Delimiter = ;
+Input Columns = skip,Date,skip,skip,Payee,Memo,skip,Inflow,skip,skip,skip
+Date Format = %d.%m.%y
+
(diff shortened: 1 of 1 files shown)


nocalla:
Thanks for the contribution!

PR #445, opened 2023-09-25
Add de paypal

martenbiehl:
**New Pull Request**

**Reference Issue:**
None

**Description**
Added German Paypal export format. The downloaded file is always called Download.CSV (capital file type, important for case-sensitive OS). I always rename the files according to the month and day, so I use "[0-9]{6}_Download.CSV". Do you generally prefer the original file name of the download?

Config changes only


nocalla:
There's a merge conflict in this one, can you base it off the most recent version of develop and try again please? 
Regarding the filename, please leave it as the original file provided by the bank, so it works for anyone else using the same bank.

PR #446, opened 2023-09-26
added de paypal

martenbiehl:
Picked the wrong branch as target in my last pull request. now it should work.

**New Pull Request**

**Reference Issue:**

**Description**
Added German Paypal export format. The downloaded file is always called Download.CSV (capital file type, important for case-sensitive OS). I always rename the files according to the month and day, so I use "[0-9]{6}_Download.CSV", but kept the file name as it is downloaded.

Config changes only


nocalla:
Thanks for the contribution!

PR #449, opened 2023-11-01
fix file matching for DE Deutsche Kreditbank checking new

ekeih:
**New Pull Request**
Relates #432 

**Description**
https://github.com/bank2ynab/bank2ynab/pull/444 added parsing for "DE Deutsche Kreditbank checking new" assuming that the filename would always include "Girokonto" in its name. But users can rename their bank accounts in the webinterface and the custom name is used to name the files. So everything between "Umsatzliste_" and "_DE" is a user controlled string, whitespace characters are replaced with "-" during the export.


nocalla:
Thanks for the contribution!

PR #453, opened 2024-02-20
Update BE BNP Paribas Fortis config to avoid match with NL Rabobank

### LINKED ISSUE #452: filename pattern of "BE BNP Paribas Fortis Export” matches “NL Rabobank-2018” files
The filename pattern of "BE BNP Paribas Fortis Export” matches “NL Rabobank-2018” files. This leads to parser errors for Rabobank csv files.

[BE BNP Paribas Fortis Export]
Source Filename Pattern = CSV_*

[NL Rabobank-2018]
Source Filename Pattern = CSV_A_20[0-9]{6}_[0-9]{6}

Since the BE filename pattern is evaluated before NL, we see parsing errors.

Solution 1: be more specific with the BE filename pattern
Solution 2: move the NL Rabobank-2018 entry up

wooter:
**New Pull Request**

**Reference Issue:**
Fixes #452

**Description**
More specific regex to catch only BE BNP Paribas Fortis CSV files

More specific regex to catch only BE BNP Paribas Fortis CSV files

### DIFF
--- a/bank2ynab.conf
+++ b/bank2ynab.conf
@@ -144,10 +144,10 @@ Input Columns = skip,Date,skip,Inflow,skip,skip,Memo,skip
 Date Format =%d/%m/%Y
 
 [BE BNP Paribas Fortis Export]
-# CSV_*.csv
+# CSV_YYYY-MM-DD-HH.MM.csv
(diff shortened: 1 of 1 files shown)


nocalla:
Thanks! That was fast!

PR #459, opened 2024-04-20
Add support for UOB SG

### LINKED ISSUE #458: UOB SG
# Savings Account

**New bank format**
*If you're submitting something else, skip or remove this section and continue below.*
- Full name of the bank: UOB
- Country (2-character country code): SG
- Any special requirements:

*Bank format: only fill out these lines if you're comfortable with it, otherwise skip:*
*To be filled in as required. Delete fields that match default values.*
```
Source Filename Pattern = ACC_TXN_History\d{14}
Source Filename Extension = .xls
Use Regex for Filename = True
Source CSV Delimiter =
Header Rows = 8
Footer Rows = 0
Input Columns = Transaction Date,Transaction Description,Withdrawal,Deposit,Available Balance
Date Format = %d %b %Y
# (see http://strftime.org/ for date format strings)
Inflow or Outflow Indicator = Separate columns
Output Columns = Date,Payee,Category,Memo,Outflow,Inflow
Output Filename Prefix = fixed_
Plugin =
```

# Credit Card Balance

**New bank format**
*If you're submitting something else, skip or remove this section and continue below.*
- Full name of the bank: UOB
- Country (2-character country code): SG
- Any special requirements: If you see a row that contains "Previous Balance", ignore it!

*Bank format: only fill out these lines if you're comfortable with it, otherwise skip:*
*To be filled in as required. Delete fields that match default values.*
```
Source Filename Pattern = CC_TXN_History\d{14}
Source Filename Extension = .xls
Use Regex for Filename = True
Source CSV Delimiter =
Header Rows = 10
Footer Rows = 0
Input Columns = Transaction Date,Posting Date,Description,Foreign Currency Type,Transaction Amount(Foreign),Local Currency Type,Transaction Amount(Local)
Date Format = %d %b %Y
# (see http://strftime.org/ for date format strings)
Inflow or Outflow Indicator = Negative sign
Output Columns = Date,Payee,Category,Memo,Outflow,Inflow
Output Filename Prefix = fixed_
Plugin =
```

limdingwen:
**New Pull Request**

**Reference Issue:** Closes #458
*Type # followed by the issue number this pull request pertains to*

**Description**
Add support for Standard Chartered SG. As part of this, adds XLS to CSV conversion support.

**Point of discussion:** SG SC has "Running Balance" as part of its export, which could be useful for YNABers reconciling. However, I left it out because I think it might be too confusing (inconsistent with other banks).

Note that trying to set a primarily number or price column to Memo or Payee results in an error in `dataframe_handler.clean_strings`.

It may be fixed by: `    modified_string_series = string_series.astype("string")`, but this was not included in this PR as it was not needed.

*Note that any submitted code must conform to the standards set by the Black linter. Please review the automated Linting test results on the pull request after saving and make any required changes or the submitted code will not be accepted.*


limdingwen:
Closes #458

bank2ynab.conf
```diff
@@ -910,6 +910,38 @@ Header Rows = 0
 Input Columns = Date,skip,Outflow,Inflow,Payee,Memo,skip
 Date Format = %d %b %Y
 
+[SG Standard Chartered Savings Account]
+# Issue: 458
+Source Filename Pattern = ACC_TXN_History_\d{14}
+Source Filename Extension = .xls
+Use Regex for Filename = True
+Source CSV Delimiter = ,
```
nocalla, line 918: You can omit entries that are the same as the default values to simplify the config, e.g. `Source CSV Delimiter = ,`. Check the defaults at the top of the file.

nocalla (changes requested):
This is great, thanks for doing the XLS plugin. I've reviewed the code and made a few suggestions - please take a look when you get a moment. 

Regarding your comments:

- Running balance isn't supported on import by YNAB unfortunately, so it's one that has to get dropped
- Your point about numerical Memo fields is an interesting one, you're probably correct in how it could be fixed, but we haven't had any reports of errors on this front to date.

limdingwen:
Thanks! I've updated the code and pushed an amended version. Note that the bank's name has been corrected.

nocalla (approved):

nocalla:
Great, thanks for updating!

PR #463, opened 2024-06-13
Update Handelsbanken SE format

JoeyPriceless:
**New Pull Request**

**Reference Issue:**
No reference issue exists.

**Description**
Updates Handelbanken SE to their new format.

Handelsbanken has updated their export format to use regular Excel, rather than their strange HTML format disguised as Excel. So we can remove the Handelsbanken plugin and just use the xls plugin.


nocalla:
Great, nice contribution - thanks!

PR #501, opened 2026-05-10
Add payee rename/transformation map config support (#404)

### LINKED ISSUE #404: Transformation map for payee names
**Other feature request**

Hey all, I am already a big fan of this script. Thanks.

One problem... let's call it nuisance... I came across is the fact that many of my recurring payees have names like "PAYPAL (EUROPE) S.A.R.L ET CIE,S.C.A." which I would like to rename. Is this a feature you would accept a Pull Request for? A decision I need your guidance with would certainly be where to provide this transformation list. Is a new section in `user_configuration.conf` okay with you or will it collide with the banks?

(Btw ever consider to transition the config files to yaml?)

nocalla:
## Summary

Adds a payee rename/transformation map so users can rewrite raw bank payee strings to friendly names via config, without touching code.

- New optional `[BankName payee_mappings]` section in each bank's `.conf` file. Each entry is `raw substring = Friendly Name`.
- Matching is case-insensitive substring; first match wins.
- `ConfigHandler` reads the section into a `payee_mappings` dict alongside other bank config.
- New `apply_payee_mappings()` function in `dataframe_handler.py`, applied after payee extraction but before `clean_strings()`.
- `BankHandler` passes the mappings dict through to the dataframe handler.
- `bank2ynab.conf` documents the new section with a commented example.
- 5 unit tests: match replaces, no-match unchanged, case-insensitive, first-match-wins, empty dict no-op.

Example usage in a bank conf file:
```
[My Bank payee_mappings]
PAYPAL (EUROPE) S.A.R.L ET CIE,S.C.A. = PayPal
AMAZON EU = Amazon
```

Closes #404.

### DIFF
--- a/bank2ynab/data/bank2ynab.conf
+++ b/bank2ynab/data/bank2ynab.conf
@@ -42,6 +42,15 @@ YNAB Account ID =
 # Just make a new [section heading] and copy only the lines from above that need a non-default value.
 # Please keep this sorted alphabetically by [country + bank name] for easier maintenance.
 #####################################################################################################
+# PAYEE RENAME MAPPINGS:
+# To rename payees for a bank, add a companion section named "[BankName payee_mappings]".
+# Each line is:  raw string to match = Friendly Name
+# Matching is case-insensitive substring; first match wins.
+# Example:
+#   [My Bank payee_mappings]
+#   PAYPAL (EUROPE) S.A.R.L ET CIE,S.C.A. = PayPal
+#   AMAZON EU = Amazon
+#####################################################################################################
 # DATE FORMATS:
 # %y   2-digit year
 # %Y   4-digit year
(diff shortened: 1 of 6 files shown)


bank2ynab/data/bank2ynab.conf
```diff
@@ -42,6 +42,15 @@ YNAB Account ID =
 # Just make a new [section heading] and copy only the lines from above that need a non-default value.
 # Please keep this sorted alphabetically by [country + bank name] for easier maintenance.
 #####################################################################################################
+# PAYEE RENAME MAPPINGS:
```
nocalla, line 48: I'm not convinced this is the best approach, as we're creating a bunch of extra config sections to account for the mappings. I would rather create a separate payee mapping conf file, now that I consider it further.

PR #523, opened 2026-06-21
Dynamic Python test matrix via endoflife.date

### LINKED ISSUE #518: Dynamic Python test matrix via endoflife.date
## Description

Updates \`testing.yml\` to always test against the three latest non-EOL Python versions without requiring manual updates when new Python versions ship. A matrix-generator job queries \`https://endoflife.date/api/python.json\`, filters to non-EOL releases, sorts descending, takes the top 3, and outputs a JSON array that the test jobs consume via \`fromJson(needs.generate-matrix.outputs.versions)\`.

Replaces the hardcoded \`python-version: ["3.12", "3.13", "3.14"]\` matrix. Currently resolves to the same three versions, but will automatically include 3.15 when it ships without any manual PR.

Also replaces any hardcoded single Python version pins in other workflows (e.g. \`codestyle.yml\`) with \`"3.x"\` (latest stable).

## Acceptance Criteria

- [ ] \`testing.yml\` has a \`generate-matrix\` job that:
  - Runs on \`ubuntu-latest\` with \`python-version: "3.x"\`
  - Calls \`https://endoflife.date/api/python.json\` with a 5-second timeout
  - Filters to cycles where \`eol\` is \`false\` or a future date
  - Sorts by version descending, takes the top 3
  - Outputs a JSON array string (e.g. \`["3.14", "3.13", "3.12"]\`) as a job output named \`versions\`
  - Exits with a clear error message if the API call fails or returns fewer than 3 versions
- [ ] The \`check\` job declares \`needs: generate-matrix\` and uses \`matrix: python-version: ${{ fromJson(needs.generate-matrix.outputs.versions) }}\`
- [ ] The hardcoded \`["3.12", "3.13", "3.14"]\` is removed from \`testing.yml\`
- [ ] \`codestyle.yml\` uses \`python-version: "3.x"\` wherever a single Python version is pinned
- [ ] CI passes on a push to \`develop\` with the dynamic matrix resolving to at least 3 versions
- [ ] \`uv run python -m pytest\` passes locally

## Testing Notes

The matrix generator can be tested locally by running the Python snippet directly:

\`\`\`python
import urllib.request, json, datetime
data = json.loads(urllib.request.urlopen(
    "https://endoflife.date/api/python.json", timeout=5
).read())
today = datetime.date.today().isoformat()
active = sorted(
    [c for c in data if str(c["eol"]) >= today],
    key=lambda c: c["cycle"], reverse=True
)[:3]
print([c["cycle"] for c in active])
\`\`\`

Verify the workflow YAML is syntactically valid before pushing. The CI run itself is the definitive test — check that the matrix shows three dynamic versions in the job summary.

nocalla:
## Summary
- Adds a `generate-matrix` job to `testing.yml` that queries endoflife.date and outputs the top 3 non-EOL Python versions
- Removes hardcoded `["3.12", "3.13", "3.14"]` matrix; new matrix auto-updates when Python versions ship or EOL
- `codestyle.yml` requires no change — it already uses `uvx` without a pinned Python version

## Test plan
- [ ] Trigger the workflow and confirm the matrix job shows three dynamic versions in the job summary
- [ ] Confirm no hardcoded version strings remain in `testing.yml`
- [ ] Confirm `codestyle.yml` uses no pinned Python version

Closes #518

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -18,12 +18,47 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  generate-matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      versions: ${{ steps.matrix.outputs.versions }}
+    steps:
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.x"
+      - name: Generate Python version matrix
+        id: matrix
+        run: |
+          python3 << 'PYEOF'
+          import urllib.request, json, datetime, sys, os
+          try:
+              data = json.loads(urllib.request.urlopen(
+                  "https://endoflife.date/api/python.json", timeout=5
+              ).read())
+          except Exception as e:
+              print(f"ERROR: Failed to fetch EOL data: {e}", file=sys.stderr)
+              sys.exit(1)
+          today = datetime.date.today().isoformat()
+          active = sorted(
+              [c for c in data if str(c["eol"]) >= today],
+              key=lambda c: c["cycle"], reverse=True
+          )[:3]
+          if len(active) < 3:
+              print(f"ERROR: Only {len(active)} active versions found", file=sys.stderr)
+              sys.exit(1)
+          versions = json.dumps([c["cycle"] for c in active])
+          print(f"Resolved matrix: {versions}")
+          with open(os.environ["GITHUB_OUTPUT"], "a") as f:
+              f.write(f"versions={versions}\n")
+          PYEOF
+
   check:
+    needs: generate-matrix
     runs-on: ${{matrix.os}}
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ["3.12", "3.13", "3.14"]
+        python-version: ${{ fromJson(needs.generate-matrix.outputs.versions) }}
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
       fail-fast: false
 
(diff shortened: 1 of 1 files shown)


nocalla:
## Code Review

**Overview:** Adds a `generate-matrix` job that queries `endoflife.date` and outputs the three newest non-EOL Python versions, replacing the hardcoded list.

### Issues

- **Lexicographic sort is fragile (latent bug):** The sort `key=lambda c: c["cycle"]` compares version strings as text. `"3.9" > "3.10"` lexicographically (since `"9" > "1"`), so the wrong three versions could be selected if a single-digit minor version is still active. Today this is hidden because Python 3.9 is already EOL, but it will surface at the next boundary. Fix with:
  ```python
  key=lambda c: tuple(int(x) for x in c["cycle"].split("."))
  ```

- **No fallback on API failure:** A transient outage of `endoflife.date` will fail every test run via `sys.exit(1)`. A hardcoded fallback (e.g. `["3.12", "3.13", "3.14"]`) on exception would keep CI resilient.

- **Extra job on the critical path:** Every push now waits for `generate-matrix` before any tests start. Worth acknowledging as a trade-off; not a blocker.

### Overall
Good idea; the sort bug is a must-fix before this is reliable.

Issue #527, opened 2026-06-28
Minus sign U+2212 (−) not recognized and being treated as an inflow in YNAB


alanlahoni:
**Describe the bug**
SE Morrow Bank has changed their minus sign in the export file from - to −
The two characters look almost identical, but they are different Unicode characters:
Hyphen-minus U+002D (-)
Minus sign U+2212: (−)

**What did you EXPECT to happen?**
I want bank2ynab to handle Minus sign U+2212: (−) in the same way as it handles Hyphen-minus U+002D (-).

**What ACTUALLY happened?**
Now bank2ynab seems to not recognize Minus sign U+2212: (−), hence the export to YNAB actually becomes positive (+) on the rows where there are minus signs and is being treated as an inflow when it is actually an outflow in YNAB.

**What did you DO? (steps to reproduce)**
1. Download Export file from Morrow Bank.
2. Run the script and all rows are becoming inflow in YNAB.

**What's your software environment?**
 - Script language: Python 3.11
 - Operating system: MacOS
 - OS version: 14.8.7
 - Installation method: uv sync

**Attachments**
Here is an example of the export file from Morrow Bank
[exportAll.csv](https://github.com/user-attachments/files/29425510/exportAll.csv)

alanlahoni:
I would really appreciate a fix for this if it is reasonably straightforward to implement. If not, I'll keep waiting. Thanks.

PR #528, opened 2026-07-14
fix: treat Unicode minus sign (U+2212) as a negative sign in monetary parsing (#527)

patchwright:
**New Pull Request**

**Reference Issue:**
#527

**Description**

SE Morrow Bank exports use the Unicode minus sign U+2212 instead of the ASCII hyphen. `_parse_monetary_string()` strips U+2212 as a stray symbol, so "−123.45" is parsed as +123.45 and every negative row is imported into YNAB as an inflow. Confirmed on HEAD 4fa82e9.

The sanitizing step in `_parse_monetary_string()` keeps only digits, periods, and the ASCII hyphen. This PR normalizes U+2212 to an ASCII hyphen before that step, so the sign survives parsing. One line of code plus a comment.

Two regression tests added to `TestParseMonetaryString`. Both fail without the fix and pass with it:

```
$ python -m pytest test/unit/test_optimisations.py -k unicode_minus -q   # before fix
FAILED test/unit/test_optimisations.py::TestParseMonetaryString::test_unicode_minus_sign
FAILED test/unit/test_optimisations.py::TestParseMonetaryString::test_unicode_minus_sign_with_thousands

$ python -m pytest test/unit/test_optimisations.py -k unicode_minus -q   # after fix
2 passed, 33 deselected

$ python -m pytest test/ -q
66 passed, 16 skipped, 29 subtests passed
```

`ruff check . --select F,E9,I` passes.

No behavior change for values that do not contain U+2212.

---
Assisted-by: Claude (code generation, reviewed and tested locally)

### DIFF
--- a/test/unit/test_optimisations.py
+++ b/test/unit/test_optimisations.py
@@ -103,6 +103,13 @@ class TestParseMonetaryString:
     def test_negative_value(self):
         assert _parse_monetary_string("-42.50") == pytest.approx(-42.50)
 
+    def test_unicode_minus_sign(self):
+        """U+2212 minus sign is parsed as negative, not stripped (#527)."""
+        assert _parse_monetary_string("−123.45") == pytest.approx(-123.45)
+
+    def test_unicode_minus_sign_with_thousands(self):
+        assert _parse_monetary_string("−1.234,56") == pytest.approx(-1234.56)
+
     def test_integer_string(self):
         assert _parse_monetary_string("100") == pytest.approx(100.0)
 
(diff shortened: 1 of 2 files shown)


test/unit/test_optimisations.py
```diff
@@ -103,6 +103,13 @@ def test_plain_decimal(self):
     def test_negative_value(self):
         assert _parse_monetary_string("-42.50") == pytest.approx(-42.50)
 
+    def test_unicode_minus_sign(self):
+        """U+2212 minus sign is parsed as negative, not stripped (#527)."""
```
nocalla, line 107: Can you remove the (#527) here please? It's just clutter in the actual code.

nocalla (changes requested):

